PR #7, opened 2018-01-19
Implement Alert

klzns:


src/components/Alert/README.md
```diff
@@ -0,0 +1,55 @@
+Alert
```
kevinch: No need for this title here. It will duplicate like so:
screen shot 2018-01-19 at 16 57 38

The "Button" before "Default"

PR #13, opened 2018-02-16
Update spinner and add copy css files

kevinch:
Spinner gained css animation and Styleguide now copies css files to lib folder.

### DIFF
--- a/src/components/Spinner/README.md
+++ b/src/components/Spinner/README.md
@@ -6,7 +6,15 @@
 </table>
 ```
 
+#### Secondary
+
+```js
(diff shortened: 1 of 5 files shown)


src/components/Spinner/README.md
```diff
@@ -6,7 +6,15 @@
...
+```js
+<table style={{ height: '40px', width: '40px', background: '#368df7' }}>
+  <Spinner secondary />
+</table>
+```
+
 _Notes:_
 
-* The `table` is only used to contain the svg for the demo.
+* The `table` are only used to contain the svg for the demo.
```
leticiawanderley: Why? The `table` is singular.

PR #25, opened 2018-02-26
Update import statement for Jest

kevinch:


src/components/Icon/index.js
```diff
@@ -13,9 +13,10 @@ export default class Icon extends Component {
 
   componentDidMount() {
     this.mounted = true
-    import(`./${this.toTitleCase(this.props.type)}`).then(
-      module => this.mounted && this.setState({ module: module.default })
-    )
+    process.env.NODE_ENV !== 'test' &&
```
klzns: Ao inves de fazer isso, usar isso:
https://github.com/facebook/create-react-app/blob/next/packages/babel-preset-react-app/index.js#L135

PR #27, opened 2018-03-02
Fix Alerts close button

cmdalbem:


gabrielgalc:
Add "flex self-center" to the div around the icon so it will be aligned in the center.

src/components/Alert/index.js
```diff
@@ -70,7 +63,7 @@ class Alert extends Component {
       <div className={`flex justify-between ${classes}`}>
         <div className="flex items-center">
           {showIcon && (
-            <Icon type={icon} fill={color} height={height} width={width} />
+            <Icon type={icon} fill={color} height={18} width={18} />
```
kevinch: This fixes the width and height of the element. We could maybe set those as defaults rather than this.

PR #29, opened 2018-03-02
Remove pointer class from input

gabrielgalc:
Hover on input brings the wrong type of cursor. Removing the class "pointer" fixes it.


cmdalbem:
nice one @gabrielgalc. I was scratching my head today why it was like that.

PR #30, opened 2018-03-02
Add tooltip component

kevinch:
screen shot 2018-03-02 at 14 49 57


cmdalbem:
Nice addition @kevinch!
I see that you exposed all the libs properties in our wrapper. I think it'd be interesting to limit a little our developers' freedom here, for the sake of usage simplicity and consistency across the Styleguide uses.
What do you think?

kevinch:
Sure thing, we can discuss this the week.

src/components/ToolTip/index.js
```diff
@@ -0,0 +1,70 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+
+import { Tooltip } from 'react-tippy'
+import 'react-tippy/dist/tippy.css'
+import './index.css'
```
klzns: Isso nao vai funcionar.

src/components/ToolTip/index.js
```diff
@@ -0,0 +1,69 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+
+import { Tooltip } from 'react-tippy'
+import 'react-tippy/dist/tippy.css'
```
klzns: Import de qualquer coisa que nao seja JS ou JSON nao funciona em Node.

PR #32, opened 2018-03-08
Improve Input API

klzns:


src/components/Input/index.js
```diff
@@ -82,19 +96,22 @@ class Input extends Component {
...
-  error: false,
   errorMessage: '',
 }
 
 Input.propTypes = {
+  id: PropTypes.string,
+  value: PropTypes.string,
+  placeholder: PropTypes.string,
+  type: PropTypes.string,
+  step: PropTypes.string,
```
kevinch: Add comments to Proptypes to help understand their meaning. Ex: https://github.com/vtex/styleguide/blob/feature/tooltip/src/components/ToolTip/index.js#L56

PR #39, opened 2018-03-16
change option input format to an object which accepts a label and a v…

brunaaleixo:
…alue (identifier)


augustocb:
@brunaaleixo refiz na branch feature/dropdown. vou vou excluir o PR!

PR #45, opened 2018-03-19
Change icons to be static components

klzns:
Remove dynamic import from icon components and make them all static, simplify icon props.

Before:
```js
import Icon from '@vtex/styleguide/lib/Icon'

<Icon type="close" width={16} height={16} />
<Icon type="check" fill="fff" />
```

This PR:
```js
import Close from '@vtex/styleguide/lib/icon/Close'
import Check from '@vtex/styleguide/lib/icon/Check'

<Check size={16} />
<Close color="#fff" />
```

Fix Toggle icon


src/components/icon/Check/index.js
```diff
@@ -20,6 +21,12 @@ class Check extends PureComponent {
   }
 }
 
+Check.defaultProps = {
+  fill: config.colors['serious-black'],
+  width: 16,
```
kevinch: Será que isso não poderia vir do config do vtex-tachyons?

PR #74, opened 2018-03-28
Refactor/size prop

lbebber:
Adds `size` prop to the components `Button`, `Input`,  and `Dropdown`, and removes the props `large` and `xLarge` from them. 

Prev: `<Button large />`
Now: `<Button size='large' />`

**This is a breaking change**, since it changes the API, although the functionality is the same.

Valid values for `size` are: `regular`, `large`, and `x-large`.

### DIFF
--- a/src/components/Button/index.js
+++ b/src/components/Button/index.js
@@ -1,5 +1,6 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
+import { SIZES } from '../../constants'
 
 class Button extends Component {
   handleClick = event => {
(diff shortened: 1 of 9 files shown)


src/components/Button/index.js
```diff
@@ -105,10 +102,8 @@ Button.defaultProps = {
 }
 
 Button.propTypes = {
-  /** Large style */
-  large: PropTypes.bool,
-  /** xLarge style */
-  xLarge: PropTypes.bool,
+  /** Button size (`regular`, `large`, or `x-large`) */
```
klzns: A documentação gerada já não preenche esses valores?

lbebber: @klzns Não (infelizmente), só aparece `SIZES`.

klzns: Bummer :(

PR #88, opened 2018-04-13
Add documentation to Toggle component. Update documentation on other …

cmdalbem:
…parts as well.


src/components/Toggle/README.md
```diff
@@ -1,39 +1,54 @@
...
 ```
 
 Semantic
+--
+
+The field this toggle controls has more meaning than just on/off: it carries an inherent right/wrong judgement of its controlled value.
+
+Use this toggle if turning it ON is highly recommended from your application standpoint. In doubt, use the Default one.
+
+Example: when configuring Security settings, you might want to signal to the user that turning ON double-authentication feature is highly recommended.
```
klzns: Aaaaahhh entendi

PR #94, opened 2018-04-20
Add react-tooltip

kevinch:
screen shot 2018-04-20 at 09 50 17

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [2.0.0-rc.40] - 2018-04-20
+
+### Added
+
+* **Tooltip**
+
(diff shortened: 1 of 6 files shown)


lbebber:
Underline breaks on linebreak, and breaks line incorrectly—both because of `display: inline-block`.
kapture 2018-04-20 at 14 04 03

kevinch:
@lbebber isso não é tão importante pois é só para a demo.

kevinch:
Style of the trigger should be free and not set. It could be a piece of text, icon etc.

lbebber:
@kevinch Makes sense. I still think it'd be worth it creating components like I said using this one, but that can be done later.

kevinch:
### react-tooltip
- use of `data` attributes
- complex, trigger and tip are two separate elements
- trigger must be `block` or `inline-block` to center the tip

### react-tippy 
- external css to be imported
- `div` as child (possibly `<p>`)

### rc-tooltip
- external css to be imported

PR #128, opened 2018-05-25
Fix/card

analuizamtg:
**What is the purpose of this pull request?**
- Replace prop `fullWidth` with `noPadding`
- Remove padding when `noPadding` is true


src/components/Card/index.js
```diff
@@ -20,8 +20,8 @@ class Card extends PureComponent {
 Card.propTypes = {
   /** Content of the card */
   children: PropTypes.node.isRequired,
-  /** Use the full width of the card. Respect the default padding (`pl6 pr6`) for elements that do not fill the full width. */
-  fullWidth: PropTypes.bool,
+  /** Use the full size of the card, without any padding. */
```
klzns: Acho que nao precisa do trecho "without any padding"

analuizamtg:
closes https://github.com/vtex/styleguide/issues/103

PR #132, opened 2018-05-28
Fix/radio click area

leticiawanderley:
Radio area was smaller than its containing `div`.


src/components/Radio/index.js
```diff
@@ -60,6 +60,10 @@ class Radio extends PureComponent {
           onChange={e => onChange(e, e.target.value)}
```
klzns: Esse onChange ta errado. Pode corrigir no PR tambem?

```diff
- onChange={e => onChange(e, e.target.value)}
+ onChange={onChange}
```

Issue #153, opened 2018-06-05
[Radio] Label should support PropTypes.node


lbebber:

lbebber:
Fixed by https://github.com/vtex/styleguide/pull/157

PR #157, opened 2018-06-11
Allows node as radio labels

lbebber:
Title is self explanatory.

Use case:
screen shot 2018-06-11 at 15 35 50


lbebber:
Fixes https://github.com/vtex/styleguide/issues/153

Issue #145, opened 2018-05-31
Simple page header


cmdalbem:
Here's a proposition of generic, simple page header to use in our admins.
(it's already implemented in Credit Control, so it's just a matter of extracting it :)

![image](https://user-images.githubusercontent.com/467471/40809792-11788a22-6502-11e8-9dc6-1016b44bde80.png)

What problem it solves
===
Our new admin sidebar provides a global solution for 1st-level navigation (between modules) and 2nd-level navigation (sections inside a module), but 3rd-level navigation is left for each module team to handle.
To improve consistency in the platform I propose this generic solution for handling simple 3rd-level navigation needs.

Design decisions
===

In the old Admin we used to deal with this with breadcrumbs, which might end up encouraging designers to create profound information architectures.
In the spirit of the Simplicity Principle, this solution encourages modules to not going much deeper in the navigation given it doesn't provide a breadcrumb or tabs, but a simple `back` link in the top.

It's important that in the implementation the vertical space for the back link is left blank if the link is not available. In other words, the main heading should always have the top margin in relation to the module available viewport. 

Examples of application
===

![image](https://user-images.githubusercontent.com/467471/40810131-3cb78548-6503-11e8-94fa-6130e5935d68.png)

![image](https://user-images.githubusercontent.com/467471/40810134-43ff51f0-6503-11e8-8a65-1be74107fd5b.png)

Figma link
===
https://www.figma.com/file/a94lX91ZmYGIDBxvb8shwLn8/VTEX-Design?node-id=191%3A1152

kevinch:
Done.

PR #198, opened 2018-07-04
Add pageheader

kevinch:
screen shot 2018-07-04 at 11 35 21

Liked to https://github.com/vtex/styleguide/issues/145


CHANGELOG.md
```diff
@@ -7,7 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
-## [5.1.0] - 2018-7-3
+## [5.2.0] - 2018-7-04
```
klzns: Faltou um zero

react/components/PageHeader/index.js
```diff
@@ -0,0 +1,57 @@
...
+import Button from '../Button'
+
+class PageHeader extends PureComponent {
+  handleClick = e => {
+    this.props.backClick && this.props.backClick(e)
+  }
+
+  render() {
+    const { backLabel, backUrl } = this.props
+    const backButton = backLabel && backUrl
```
klzns: backUrl nao ta sendo usado, é isso msm?

react/components/PageHeader/README.md
```diff
@@ -0,0 +1,18 @@
...
+<PageHeader title="Invoices" />
+```
+
+With back link
+
+```js
+<PageHeader
+  title="Invoice"
+  backLabel="Back"
+  backUrl="javascript:;"
```
klzns: Podia realmente ser uma URL pro exemplo

react/components/PageHeader/index.js
kevinch: True, não usei afinal.

cmdalbem:
There are a couple of problems in the implementation that deviate from the original design:

- [ ] Background color should be light-silver, not near-white (we're killing near-white because of contrast issues)
- [ ] Margins are a little off, should be like this:
image
- [ ] Although the back link could receive a label "Back", it would be more interesting if it was more semantic label, like in the example use cases from the [original issue](https://github.com/vtex/styleguide/issues/145). I think it'd be better not to use "Back" in the Styleguide examples.
- [ ] It'd be interesting if the wether the back links is used or not, the component height and inner margins would stay the same. When overlapping the screenshots of the implemented version we can see that both the height and the margins change:
image

cmdalbem:
There are a couple of problems in the implementation that deviate from the original design. I documented them here: https://github.com/vtex/styleguide/issues/199

PR #261, opened 2018-08-15
Change badge font size to .75rem

pedroteosousa:
### Before
image
### After
image


klzns:
A @ntmedina disse que voltaram atras dessa mudança

PR #265, opened 2018-08-17
Fix/icon block

### LINKED ISSUE #254: Add display:block to icons
Our icons (`<svg>` elements) are by default inline items, which means that they always render a small padding below them, throwing off alignment in some cases.

Adding an option to display them as blocks would be very convenient.

akaFTS:
Fix #254


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 - Add CSS classes to all icon components
+- Add option to display icons as block
```
klzns: O que acha de "Add prop `block` to display icons as block"

PR #299, opened 2018-08-22
Add missing tokens

klzns:


react/components/NumericStepper/index.js
```diff
@@ -215,7 +213,7 @@ export default class NumericStepper extends React.Component {
             >
               <span className="b">
                 {/* fullwidth hyphen-minus (U+FF0D) http://graphemica.com/%EF%BC%8D */}
-                &#xFF0D;
+                －
```
kevinch: Any chance those signs could in a constant file to avoid error in the future?

PR #315, opened 2018-09-01
possibility to hide close button on modal

marcelotokarnia:


react/components/Modal/index.js
```diff
@@ -66,6 +69,8 @@ Modal.propTypes = {
   closeOnEsc: PropTypes.bool,
   /** Close the modal on overlay click (default true) */
   closeOnOverlayClick: PropTypes.bool,
+  /** Show the close icon on upper right corner (default true) */
+  showCloseIcon: PropTypes.bool,
```
kevinch: should be `hideCloseIcon`if default is `true`.

.eslintrc
```diff
@@ -1,3 +1,27 @@
 {
-  "extends": ["vtex-react-app"]
+  "parser": "babel-eslint",
```
kevinch: Why update this file?

marcelotokarnia: This is the IO linting rules. 
Soon it will be automatic and it will not be up to the user anymore to define his linting rules. So better get used to it ASAP.

react/components/Modal/index.js
marcelotokarnia: I copied the same name that 'react-responsive-modal' uses. I think it's better to use names the devs are used to instead of mapping same functionalities to different names.

I think it is better to leave it that way. What do you think ?

PR #317, opened 2018-09-06
Feature/pageheader side children

kevinch:
Let devs use the side of the header to add any children.


kevinch:
ref https://github.com/vtex/styleguide/issues/307

Issue #144, opened 2018-05-31
Simple Search Input


cmdalbem:
Here's a simple search field input we use in Credit Control.
(we have a working version of it in Credit Control, but it has some _gambiarras_ so maybe it's better to start from scratch)

![image](https://user-images.githubusercontent.com/467471/40809607-6e247156-6501-11e8-92aa-2ce2f0fcea7e.png)

What problem it solves
===
Just a simple text input for simple searching needs.
If a search is "on" and the user hasn't changed the input to search for something different, the search button becomes clear button for clearing the app state.

Design decisions
===

- The `search` button doesn't have a text label for aesthetic reasons. We think this could have had negative effects on usability if done a some years ago, however from our benchmark research this pattern it totally common place in the Internet nowadays.
- This component is for active searching. If your app has something like a preemptive or dynamic searching (you show results while the user is typing) maybe the search button could be disabled.

Examples of application
===

![image](https://user-images.githubusercontent.com/467471/40809289-5cf83f44-6500-11e8-9e17-44b73649e623.png)

![image](https://user-images.githubusercontent.com/467471/40809307-6749a442-6500-11e8-902d-e674502b2ac0.png)

Benchmark
===
Carbon: http://www.carbondesignsystem.com/components/search/code

Figma link
===
https://www.figma.com/file/a94lX91ZmYGIDBxvb8shwLn8/VTEX-Design?node-id=148%3A969

klzns:
Started development of this feature at:
https://github.com/vtex/styleguide/commits/feature/search-input

kevinch:
@cmdalbem the submit action can only happen on `enter` keypress?

cmdalbem:
@kevinch não, pode ser no clique também! Por isso que ele está à direita e com cor de ação. Se o submit fosse só no `enter` o ícone ficaria melhor à esquerda e com menor proeminência.
O comportamento desse componente estava perfeito no Credit Control, pode usar ele como referência ;)

cmdalbem:
Closed by https://github.com/vtex/styleguide/pull/281

PR #332, opened 2018-09-14
Improve input Search

ntmedina:
Context: We don't need to use blue color (our interactive color) in the search icon because it isn't an interactive one, it exists just for semantic purposes.

- Fix search icons (Color and shapes)
___

Antes
screen shot 2018-09-14 at 3 17 47 pm

___

Depois
screen shot 2018-09-14 at 3 16 15 pm


react/components/InputSearch/index.js
```diff
@@ -2,7 +2,7 @@ import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import Input from '../Input'
 import SearchIcon from '../icon/Search'
-import DenyIcon from '../icon/Deny'
+import FailureIcon from '../icon/Failure'
```
lbebber: 👀 acho que deveriamos renomear isso (ou criar um alias)

gabrielgalc:
Looks better to me!

cmdalbem:
The component had been originally designed for the cenario where a submit action is needed, as documented in the original issue: https://github.com/vtex/styleguide/issues/144. In Credit Control we couldn't have a "dynamic search" behavior for technical reasons.
I suggest we review together this design to see how we can make it work for both cenarios. Please don't merge this PR since it'll effect projects currently using it.

felippenardi:
@ntmedina Is this still going to be merged?

cmdalbem:
We discussed this in a design meeting a couple of weeks ago and decided a redesign was needed. I created an issue to address these: https://github.com/vtex/styleguide/issues/423

PR #335, opened 2018-09-18
forces table re-render when key changes

marcelotokarnia:
There was a client reporting that whenever his locale changed the table wouldn't update unless he forced re-rendering somehow (e.g. changing some row content, adding/deleting row)

That does make sense, because Table is a PureComponent, and since locale is in the context, not in props and not in the state, it really was designed not to update itself automatically in this case.

So we should provide a way the user can tell the table to re-render whenever he wants. Passing through this additional prop will do this, so the client can create a hash based on whatever he wants to tell the table when to update. 

In this case I could instruct him to use the locale as a key and it would re-render as expected when the locale changed.

I tried to explain the problem so maybe we can discuss a different approach to solve the problem different than this.

Thoughts ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [6.3.3] - 2018-09-18
+
 ## [6.3.2] - 2018-09-13
 ### Added
 
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
   "vendor": "vtex",
   "name": "styleguide",
-  "version": "6.3.2",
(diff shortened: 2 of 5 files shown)


ohmyguigs:
@marcelotokarnia I guess if locale changes, the parent component would have to translate the data or even the Header labels of the column. And if you change the items or the schema, the table rerender beacause of prop change? Is there a working example he showed with code?
seems useless to me a rerender trigger, given that no other prop changed. Wouldn't a loader on parent component solve this?

ohmyguigs:
Só pra referencias futuras, vimos que o caso de uso é usar um `<FormatedMessage />`do `intl` dentro do cellRenderer no schema. Relamente a tabela nao sabe que o locale mudou e é necessário algum parametro mudar pra ela re-renderizar.

PR #344, opened 2018-09-27
Make styles and classes names props

henriquecaraujo:
Removed styles and classes name objects from component and added it to a defaultProps.


brunojdo:
I explain to @henriquecaraujo the central proposal of our styleguide, he understood, and I will drop this PR.

PR #351, opened 2018-09-28
added classes to Tabs component

eduardofidera:
added new classes to Tabs component elements, so it can be styled easily


lbebber:
merged in https://github.com/vtex/styleguide/pull/352

PR #355, opened 2018-10-01
Fix/numeric stepper label

lbebber:
Use the label tag on the NumericStepper only when there is a label. 
This prevents focusing on the input when either the increment or the decrement buttons are pressed, and thus popping up the keyboard on mobile devices, as observed on iOS.

Also makes the dev server publicly accessible, for testing on devices running on a proxy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [6.4.4] - 2018-10-01
+
 ### Fixed
 
 - **NumericStepper** Refrains from using the label tag if there is no label, to prevent the keyboard from popping up on iOS unnecessarily
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+- **NumericStepper** Refrains from using the label tag if there is no label, to prevent the keyboard from popping up on iOS unnecessarily
```
Caio-Batista: I think in this case 'Refrains the use of label tag ' is better here than  'Refrains from using'.

react/components/NumericStepper/index.js
```diff
@@ -153,8 +153,15 @@ export default class NumericStepper extends React.Component {
       'x-large': `pv5 f4 ${block ? 'flex-grow-1' : 'w4'}`,
     }
 
+    // Refrain from using label tag if not needed, to prevent
```
Caio-Batista: Ditto.

CHANGELOG.md
lbebber: As far as I know, you refrain _from sth_, not refrain _sth_

lbebber: Checkout the first example https://dictionary.cambridge.org/us/dictionary/english/refrain

Caio-Batista: Great, approved then.

PR #378, opened 2018-10-11
Enhance Table

ohmyguigs:
# What does this PR do?

 - Opinionated hover style in Table when there is onRowClick function.
 - Insert `density` as prop to Table
 - Adds Density control in toolbar (`density` prop inside toolbar shape)
 - Adds Density Icon
 - [Bonus] Makes Icon Table grid automatic in Icons Readme (now you can just add a new icon on the require object and ta magic is done automatically ✨)

### DIFF
--- a/react/components/Table/SimpleTable.js
+++ b/react/components/Table/SimpleTable.js
@@ -4,7 +4,6 @@ import { Column, Table as VirtualTable, AutoSizer } from 'react-virtualized'
 import ArrowDown from '../icon/ArrowDown'
 import ArrowUp from '../icon/ArrowUp'
 const ARROW_SIZE = 11
-const ROW_HEIGHT = 64
 const HEADER_HEIGHT = 36
 
 class SimpleTable extends PureComponent {
@@ -35,6 +34,7 @@ class SimpleTable extends PureComponent {
       sort: { sortOrder, sortedBy },
       onSort,
       updateTableKey,
+      rowHeight,
     } = this.props
     const properties = Object.keys(schema.properties)
     // hydrate items with index when 'indexColumn' prop is true
@@ -55,7 +55,7 @@ class SimpleTable extends PureComponent {
               width={width}
               height={height}
               headerHeight={HEADER_HEIGHT}
-              rowHeight={ROW_HEIGHT}
+              rowHeight={rowHeight}
               rowCount={newItems.length}
               rowGetter={({ index }) => newItems[index]}
               className="flex flex-column"
@@ -74,7 +74,7 @@ class SimpleTable extends PureComponent {
                 `flex flex-row items-center ${
                   index === -1 ? 'bt bb' : 'bb'
                 } b--muted-4 ${
-                  onRowClick && index !== -1 ? 'hover-bg-action-secondary pointer' : ''
+                  onRowClick && index !== -1 ? 'pointer hover-bg-near-white hover-c-link' : ''
                 }`
               }
             >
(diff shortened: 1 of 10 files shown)


react/components/Table/SimpleTable.js
```diff
@@ -144,41 +146,28 @@ SimpleTable.defaultProps = {
...
-  onRowMouseOut: () => {},
-  onRowMouseOver: () => {},
   sort: {
     sortOrder: null,
     sortedBy: null,
   },
 }
 
 SimpleTable.propTypes = {
-  /** Array of objects with data */
```
kevinch: Why removing those comments?

react/components/Table/Toolbar.js
```diff
@@ -108,6 +115,80 @@ class Toolbar extends PureComponent {
...
+                <div
+                  className="absolute z-999 ba b--light-gray br2 mt2"
+                  style={BOX_SHADOW_STYLE}
+                >
+                  <div
+                    className="w-100 b2 br2 bg-base"
+                    style={{ width: DENSITY_BOX_WIDTH }}>
+                    <div style={{ height: 3 * FIELDS_BOX_ITEM_HEIGHT }} className="overflow-scroll">
+                      <div
+                        className={`flex justify-between ph6 pv3 pointer hover-bg-light-silver bl bw1 ${
```
kevinch: This code is repeated. Could be a const.

react/components/Table/Toolbar.js
```diff
@@ -108,6 +115,80 @@ class Toolbar extends PureComponent {
...
+              {isDensityBoxVisible && (
+                <div
+                  className="absolute z-999 ba b--light-gray br2 mt2"
+                  style={BOX_SHADOW_STYLE}
+                >
+                  <div
+                    className="w-100 b2 br2 bg-base"
+                    style={{ width: DENSITY_BOX_WIDTH }}>
+                    <div style={{ height: 3 * FIELDS_BOX_ITEM_HEIGHT }} className="overflow-scroll">
+                      <div
```
kevinch: Maybe a `map` over those 3 options?

react/components/Table/Toolbar.js
```diff
@@ -270,6 +357,12 @@ Toolbar.propTypes = {
     inputSearch: PropTypes.shape({
       onSubmit: PropTypes.func,
     }),
+    density: PropTypes.shape({
+      label: PropTypes.string,
```
kevinch: Maybe a better name for this "label" since there are 3 more below.

react/components/icon/README.md
```diff
@@ -1,242 +1,82 @@
...
-        <ShoppingCart size={demoSize} />
-      </td>
-    </tr>
-  </tbody>
-</table>
+    return (
+      <table className="w-100">
+        <tbody>
+          {
+            chunkedIconsMatrix.map((iconsLine, row) => {
```
kevinch: hint: `.map((iconsLine, row) => <tr key={}>...`

react/components/Table/Toolbar.js
ohmyguigs: Gr8 idea

react/components/Table/Toolbar.js
ohmyguigs: is it better if I call it `button`? So they are all labels `button`, `low`, `medium` and `high`

kevinch: `buttonLabel` or `labelButton`?

kevinch: Did not see that you had renamed the labels below. Looks fine like this.

kevinch: Maybe `button` yes, or similar. Like `buttonLabel`

PR #386, opened 2018-10-18
Enhancement/table

ohmyguigs:
## What is the purpose of this PR?

 - Change table schema clone logic from componentDidUpdate to getDerivedStateFromProps
 - Add feature to toggle box align of toolbar menus

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


salesfelipe:
I don´t think that cloning the whole object is the best approach for hiding the table columns. Maybe you could save a hash table in the state, that will store the keys for the columns that won't be displayed on the table and then on the render method, you could check if that given column should be displayed or not.

cmdalbem:
@ohmyguigs I think @salesfelipe 's suggestion would also fix that problem of the columns being reordered when toggling their states, right?

PR #388, opened 2018-10-25
Add typography styles tokens

klzns:
Change all references to font-size (`.f1`, `.f2`, etc) and (most) font-weights (`.fw3`, `.fw5`, etc) in favor of typography styles tokens introduced here: https://github.com/vtex/vtex-tachyons/pull/14

Basically it was mapped like this:

Before | After
---|---
`f1` | `t-heading-1`
`f2` | `t-heading-2`
`f3` | `t-heading-3`
`f4` | `t-body`
`f5` | `t-body`
`f6` | `t-small`

Some components used `f3` and `f2` for things that were not headers, those instances were substituted with `t-body`.


lbebber:
Implemented by https://github.com/vtex/styleguide/pull/431

PR #391, opened 2018-10-31
Vertical tabs

victorsenam:
The need arises from `pages-admin` where we want to create a vertical mode switcher, like this:
screen shot 2018-10-31 at 10 49 42
This is from [this figma](https://www.figma.com/file/Fq9myz1oHXLy1Tzygb7jbiUf/Storefront-(Layout-%26-Style)?node-id=365%3A2).

I've figured that the most important thing, for now, is that it is vertical and we can maybe add a prop that allows it to use the secondary color later. There's also some rounded corner on the border created by the figma that looks awesome IMO, but that's maybe a stretch.

It looks like this on the styleguide:
screen shot 2018-10-31 at 10 52 06

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 
-- **Tabs** `direction` prop to control wether the tabs are displayed vertically or horizontally
-- **Tab** `direction` prop to control wether the border is below or to the right of the active tab
+- **Tabs** `vertical` prop to control the direction of the tabs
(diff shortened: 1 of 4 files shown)


thiagomurakami:
I think a prop `vertical (boolean)` could be a good alternative to `direction`. WDYT?

This way, the code for vertical tabs would look like:
```jsx
<Tabs vertical>
  {/* content */}
</Tabs>
```

victorsenam:
I've talked to @ntmedina and we've decided we'd better do a `VerticalTab` component, because that one would be focused on the use of icons instead of text and we could add tooltips and specific beautiful hovers that only work on icons.

Issue #393, opened 2018-10-31
NumericStepper "ghost" variation


hapoza:
## Overview
I think NumericStepper should have a variation similar to Button, following the Button's tertiary variation, where it has less importance on the UI.
This would be great for cases where there are a lot of NumericSteppers on the same screen.

## Purpose
For `vtex.product-customizer`, there's a use case for the NumericStepper where it doesn't have a lot of screen real estate. Other than It being too big, it looks quite bulky compared to the rest of the prototype for Delivery Dreamstore.

## Context
This discussion started on https://github.com/vtex-apps/product-customizer/pull/3, and it made sense that it maybe could be a feature of Styleguide.

## Examples
This is how NumericStepper is by default, it is too big for that use case on mobile screens 
![numeric-stepper-defaults](https://user-images.githubusercontent.com/27775611/47821945-3c5ada80-dd41-11e8-9d13-ac19e295860b.png)

Here's what was done and possibly how it could look:
![numeric-stepper-styled](https://user-images.githubusercontent.com/27775611/47821987-63b1a780-dd41-11e8-85b2-80bf76715bfc.png)

gabrielgalc:
Great, it makes total sense, @hapoza .

gabrielgalc:
Next step would be a PR here with the changes you made on the product customizer, right?

hapoza:
Yes, I can do that!

PR #396, opened 2018-11-05
NumericStepper lean mode

lbebber:
Adds lean mode to NumericStepper. Details of the styling might be tweaked later.

Closes https://github.com/vtex/styleguide/issues/393

lean stepper

### DIFF
--- a/react/components/NumericStepper/index.js
+++ b/react/components/NumericStepper/index.js
@@ -136,56 +136,79 @@ export default class NumericStepper extends React.Component {
 
   render() {
     const { value, displayValue } = this.state
-    const { maxValue, minValue, size, block, label } = this.props
+    const { maxValue, minValue, size, block, label, lean } = this.props
 
     const isMin = value <= normalizeMin(minValue)
     const isMax = value >= normalizeMax(maxValue)
 
-    const buttonSizeClasses = {
-      regular: 'pv3 f6',
-      large: 'pv4 f5',
-      'x-large': 'pv5 f4',
-    }
+    // TODO: Make this less cumbersome - will be done when x-large is killed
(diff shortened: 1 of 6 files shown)


react/components/NumericStepper/index.js
```diff
@@ -136,56 +136,79 @@ export default class NumericStepper extends React.Component {
...
     const isMin = value <= normalizeMin(minValue)
     const isMax = value >= normalizeMax(maxValue)
 
-    const buttonSizeClasses = {
-      regular: 'pv3 f6',
-      large: 'pv4 f5',
-      'x-large': 'pv5 f4',
-    }
+    // TODO: Make this less cumbersome - will be done when x-large is killed
+    const buttonSizeClasses = (lean
```
athoscouto: nit: when we have nested ternary operators it may be a good idea to refactor into a function with more explicit logic.
I'd leave this one as it is, tough.

athoscouto: Since it is not an object with the classes anymore, I'd call it `buttonSizeClass` instead of `buttonSizeClasses`.
Ditto bellow on `inputSizeClasses`

athoscouto: Don't know if this makes that much sense, since they are indeed classes. Maybe this name is correct now and wasn't before ¯\_(ツ)_/¯

lbebber: @athoscouto agree 100%, but this whole bit is going to be refactored soon, as `x-large` sizes are going to be removed, so I went with the easy way out for now

athoscouto:
As @brunojdo noted, would be nice to have it working on an account/namespace. I'll also try to do it on next PRs :)

PR #403, opened 2018-11-07
Fix/spinner css

lbebber:
Changes the animation of the spinner from SVG based to CSS based. This fixes a problem where the spinner wasn't playing during certain stages of the page load—namely, the beginning of it—due to how SVG animations work.

### DIFF
--- a/react/components/Spinner/index.js
+++ b/react/components/Spinner/index.js
@@ -5,58 +5,121 @@ import { baseClassname } from '../icon/utils'
 const radius = 40
 const circ = 2 * radius * Math.PI
 
-const Spinner = ({ color, size, block }) => (
-  <svg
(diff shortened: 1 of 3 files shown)


react/components/Spinner/index.js
```diff
@@ -5,53 +5,75 @@ import { baseClassname } from '../icon/utils'
...
+      <svg
+        className={`${baseClassname('spinner')} ${
+          !color ? 'c-action-primary' : ''
+        } ${block ? 'db' : ''}`}
+        xmlns="http://www.w3.org/2000/svg"
+        viewBox="0 0 100 100"
+        preserveAspectRatio="xMidYMid"
+        height={size}
+        width={size}>
+        <style>
```
klzns: que maroto

PR #467, opened 2018-12-26
Add new component Tooltip

klzns:
**Ainda falta escrever a documentação.**

image


react/components/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,31 @@
...
+      {children}
+      <div
+        className={`${styles.tooltip} ${
+          visible ? '' : 'dn'
+        } vtex-tooltip__container absolute h-auto mw5 bg-base--inverted pa1 br1 z-5`}>
+        <div className="vtex-tooltip__label bg-base--inverted c-on-base--inverted pa2 t-body">
+          {label}
+        </div>
+        <div
+          className={`${styles.arrow} vtex-tooltip__arrow c-on-base absolute`}
```
lbebber: a `c-on-base` nao vai ser necessariamente igual ao `bg-base--inverted`

lbebber: só pra ter certeza, essa seria a cor da flecha, isso?

cmdalbem:
@klzns @lbebber o que falta pra mandarmos essa?

lbebber:
The tooltip is to be rewritten—maybe with the same solution as pointed on https://github.com/vtex/styleguide/issues/505—in a way that is less disruptive regarding the DOM

cmdalbem:
@lbebber I see no point in keeping this PR open then, I'll close it.

PR #468, opened 2018-12-27
Refactor components' organization and show 1 per page

cmdalbem:
- Reorder components and groups to be ordered alphabetically
- Finally, makes every component fill only one page, no more pages for groups of components.

### DIFF
--- a/styleguide.config.js
+++ b/styleguide.config.js
@@ -144,13 +144,6 @@ module.exports = {
           ],
         },
         {
-          name: 'Display',
-          components: [
-            'react/components/Table/index.js',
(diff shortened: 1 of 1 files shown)


styleguide.config.js
```diff
@@ -77,15 +77,31 @@ module.exports = {
       name: 'Components',
       content: './docs/components.md',
       skipComponentsWithoutExample: true,
-      sectionDepth: 1,
+      sectionDepth: 2,
```
salesfelipe: why ?

cmdalbem: @salesfelipe it's for the "show 1 per page" part.

PR #470, opened 2018-12-27
Fix styles for disabled and read-only states

cmdalbem:
- Removes hover effects of `read-only` Input, since it's not actually interactive.
- Fixes type color on disabled state.

image

### DIFF
--- a/react/components/Input/README.md
+++ b/react/components/Input/README.md
@@ -60,8 +60,15 @@ class InputExamples extends React.Component {
   render() {
     return (
       <div className="w-40">
+      <div className="mb5">
+          <Input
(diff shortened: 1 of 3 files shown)


react/components/Input/README.md
```diff
@@ -60,8 +60,15 @@ class InputExamples extends React.Component {
   render() {
     return (
       <div className="w-40">
+      <div className="mb5">
+          <Input
+            value={"This input is read-only."}
```
salesfelipe: You don't need the : "{"

cmdalbem: thx

PR #509, opened 2019-01-23
Add creatable option

BrunoLerner:
It adds a flag prop that renders a different component, which is the CreatableSelect, where you can add custom options.

There is a problem with the limit of things you can select, this is what happens:
screen shot 2019-01-23 at 15 30 47

After debugging a little bit, I found out that if I remove this function getValueContainerHeightFromSize() from the styles prop it works better:
screen shot 2019-01-23 at 15 30 05

But I didn't remove the function, because I don't know if this is working good for any other use case.

### DIFF
--- a/react/components/Select/README.md
+++ b/react/components/Select/README.md
@@ -2,7 +2,7 @@
 
 ### 👍 Dos
 
-- Mind the order of the options, like putting more probable to be picked on that. In doubt, sort them alphanumerically (from A to Z and from 0 to 9).
+- Mind the order of the options, like putting the more probable one to be selected at the top. In doubt, sort them alphanumerically (from A to Z and from 0 to 9).
 
 ### 👎 Don'ts
 
--- a/react/components/Select/README.md
+++ b/react/components/Select/README.md
@@ -161,7 +161,6 @@ const options = [
     label="Label"
     options={options}
     isMulti={true}
-    errorMessage="Required!"
     onChange={values => {
       console.log(`[Select] Selected: ${JSON.stringify(values, null, 2)}`)
     }}
--- a/react/components/Select/README.md
+++ b/react/components/Select/README.md
@@ -156,7 +156,6 @@ const options = [
     label: 'Second Option',
   },
 ]
-
 ;<div>
   <Select
     label="Label"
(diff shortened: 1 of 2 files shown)


cmdalbem:
@BrunoLerner in the Examples section you use the prop as `isCreatable`, but in the code you call it `creatable`. I think it should be called `isCreatable` to match other similar props.
Please also add the documentation for the prop, like the others.

react/components/Select/index.js
```diff
@@ -32,108 +33,208 @@ const Select = ({
   placeholder,
   size,
   value,
+  creatable,
```
cmdalbem:
```suggestion
  isCreatable,
```

react/components/Select/index.js
```diff
@@ -32,108 +33,208 @@ const Select = ({
...
-      onChange={onChange}
-      options={options}
-      styles={{
-        control: style => {
-          const errorStyle = errorMessage
-            ? {
-                borderColor: COLORS.red,
-              }
-            : {}
+    {creatable ? (
```
cmdalbem:
```suggestion
    {isCreatable ? (
```

react/components/Select/index.js
```diff
@@ -198,6 +299,7 @@ Select.propTypes = {
       })
     ),
   ]),
+  creatable: PropTypes.bool,
```
cmdalbem:
```suggestion
  isCreatable: PropTypes.bool,
```

ericreis:
@cmdalbem In the select PR @klzns asked to change the `isDisabled` to `disabled` to match the API of the other _styleguide_ components. So I think the correct is clearable in this case, we need to change the others.

react/components/Select/README.md
```diff
@@ -142,6 +143,33 @@ const options = [
...
+;<div>
+  <Select
+    label="Label"
+    options={options}
+    isMulti={true}
+    errorMessage="Required!"
+    onChange={values => {
+      console.log(`[Select] Selected: ${JSON.stringify(values, null, 2)}`)
+    }}
+    isCreatable
```
cmdalbem:
```suggestion
    creatable
```

BrunoLerner:
@cmdalbem I changed from `isCreatable` to `creatable` because of what @ericreis said, and I forgot to update the example :)

PR #524, opened 2019-02-04
Feature/select input ref

ericreis:
This PR fixes a lost of focus behavior when the props of the select were updated. This case was particularly painful when trying to make async select options.

### DIFF
--- a/react/components/EXPERIMENTAL_Select/index.js
+++ b/react/components/EXPERIMENTAL_Select/index.js
@@ -23,8 +23,6 @@ class Select extends Component {
     super(props)
 
     this.inputId = `react-select-input-${uuid()}`
-      inputId: `react-select-input-${uuid()}`,
-    }
   }
 
   componentDidUpdate() {
@@ -49,8 +47,6 @@ class Select extends Component {
       size,
       value,
     } = this.props
-
-    const { inputId } = this.state
 
     const reactSelectComponentProps = {
       autoFocus,
--- a/react/components/EXPERIMENTAL_Select/index.js
+++ b/react/components/EXPERIMENTAL_Select/index.js
@@ -22,13 +22,13 @@ class Select extends Component {
   constructor(props) {
     super(props)
 
-    this.state = {
+    this.inputId = `react-select-input-${uuid()}`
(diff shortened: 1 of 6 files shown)


react/components/EXPERIMENTAL_Select/index.js
```diff
@@ -16,140 +18,159 @@ const getOptionValue = option => {
...
-          />
-        )
-      },
-      DropdownIndicator: function DropdownIndicator(props) {
-        return <DropdownIndicatorComponent size={size} {...props} />
+class Select extends Component {
+  constructor(props) {
+    super(props)
+
+    this.state = {
```
ohmyguigs: If you are not gonna manipulate (setState) then this should be just a const. it doesn't have to be a state.
```suggestion
    this.inputId = `react-select-input-${uuid()}`
```

react/components/EXPERIMENTAL_Select/index.js
```diff
@@ -16,140 +18,159 @@ const getOptionValue = option => {
...
+      inputId: `react-select-input-${uuid()}`,
+    }
+  }
+
+  componentDidUpdate() {
+    document.getElementById(this.state.inputId).focus()
```
ohmyguigs:
```suggestion
    document.getElementById(this.inputId).focus()
```

react/components/EXPERIMENTAL_Select/index.js
```diff
@@ -16,140 +18,159 @@ const getOptionValue = option => {
...
-          ...errorStyle,
-          borderWidth: '.125rem',
+      defaultValue,
+      getOptionValue,
+      isClearable: true,
+      isDisabled: disabled,
+      isLoading: loading,
+      isMulti: multi,
+      noOptionsMessage,
+      inputId: inputId,
```
ohmyguigs:
```suggestion
      inputId: this.inputId,
```

react/components/EXPERIMENTAL_Select/index.js
ohmyguigs: Or maybe it could be a `static` inside the class.

PR #533, opened 2019-02-11
Fix dropdown height problem

theomoura:
As title says

image

#### Types of changes


react/components/Dropdown/index.js
```diff
@@ -107,7 +107,7 @@ class Dropdown extends Component {
 
     let classes = 'bg-transparent bn w-100 h-100 '
     let containerClasses = `${isInline ? '' : 'bw1'} br2 relative `
-    let selectClasses = 'o-0 absolute top-0 left-0 w-100 bottom-00 '
+    let selectClasses = 'o-0 absolute top-0 left-0 w-100 h-100 bottom-00 '
```
lbebber: I think the proper solution here would be to fix the `bottom-00` class—the correct name is `bottom-0`

ohmyguigs: This problem occurs only in Windows Chrome and MS Edge.`h-100` is not necessary, just fixing the `bottom-00` typo for `bottom-0` fixes it.

```suggestion
    let selectClasses = 'o-0 absolute top-0 left-0 w-100 bottom-0 '
```

theomoura: This seems to resolve the problem. Thx @lbebber @ohmyguigs

ohmyguigs:
I'm gonna close this PR since we fixed this in 8.x and 9.x already on vtex/styleguide/pull/520

PR #534, opened 2019-02-11
[TABLE] Add isLoading prop

kevinch:
Allow table's toolbar buttons to use the `isLoading` state:

screen shot 2019-02-11 at 12 42 29

### DIFF
--- a/react/components/Table/Toolbar.js
+++ b/react/components/Table/Toolbar.js
@@ -268,6 +268,7 @@ class Toolbar extends PureComponent {
               }
               disabled={loading}
               variation="tertiary"
+              isLoading={download.isLoading}
               size="small"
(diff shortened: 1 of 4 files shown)


react/components/Table/Toolbar.js
```diff
@@ -337,6 +340,7 @@ class Toolbar extends PureComponent {
             <ButtonWithIcon
               icon={<IconPlus solid size={LIGHT_ICON_SIZE} />}
               disabled={loading}
+              isLoading={newLine.isLoading}
```
salesfelipe: Maybe you should add this to the propTypes also.

kevinch: Good, catch, updated.

Issue #515, opened 2019-01-29
Inputs props not appearing in docs


ericreis:
**Describe the bug**
The `props` for all Inputs are not appearing in the documentation

**To Reproduce**
1. Go to https://styleguide.vtex.com/#/Components/Forms/Input
2. See that there are no `props` in the documentation
3. The same occurs for the input components

**Expected behavior**
Show the component `props` in the documentation

**Screenshots**
![image](https://user-images.githubusercontent.com/5971264/51913705-4e8afa00-23be-11e9-9aab-60d387eb7979.png)

**Additional context**

PR #535, opened 2019-02-13
Input components using HOC withForwardRef

cezarguimaraes:
This PR intends to fix ref forwarding for Button/ButtonWithIcon. We also standardized other input components that needed to forward refs.

As an extra, this fixed (it seems) the documentation for some components:
- Before:
image
- After
image


ericreis:
Fixes #515

PR #537, opened 2019-02-14
Fix focus behavior when typing in the search input

ericreis:
Before this PR, we were applying focus to the input component of the **EXPERIMENTAL_Select** in the `componentDidUpdate` lifecycle. This was causing some weird focus behavior when a change in another part of the page triggers an update on the **EXPERIMENTAL_Select** and, therefore, the select was focused, even though the user was not using the component.

Now, we only focus the **EXPERIMENTAL_Select** component when is being used.

### DIFF
--- a/react/components/EXPERIMENTAL_Select/index.js
+++ b/react/components/EXPERIMENTAL_Select/index.js
@@ -22,11 +22,22 @@ class Select extends Component {
   constructor(props) {
     super(props)
 
+    this.state = {
+      searchTerm: undefined,
+    }
+
     this.inputId = `react-select-input-${uuid()}`
   }
 
-  componentDidUpdate() {
-    document.getElementById(this.inputId).focus()
+  componentDidUpdate(prevProps, prevState) {
+    const { searchTerm } = this.state
+    const { searchTerm: prevSearchTerm } = prevState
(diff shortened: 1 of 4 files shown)


ericreis:
Fixes #541

react/components/EXPERIMENTAL_Select/index.js
```diff
@@ -22,11 +22,22 @@ class Select extends Component {
...
-  componentDidUpdate() {
-    document.getElementById(this.inputId).focus()
+  componentDidUpdate(prevProps, prevState) {
+    const { searchTerm } = this.state
+    const { searchTerm: prevSearchTerm } = prevState
+    const { loading } = this.props
+    const { loading: prevLoading } = prevProps
+
+    if (searchTerm !== prevSearchTerm || loading !== prevLoading) {
+      document.getElementById(this.inputId).focus()
```
klzns: Use ref instead of `getElementById`.

Also, I think that's not a very reliable solution. 

Can you apply focus just when the user selects an option?

klzns: As discussed personally, the component don't provide ref.

Also, the searchTerm is not passed as props, that's why its kept inside as a state.

PR #538, opened 2019-02-14
Feature/filters

ohmyguigs:
### What does this PR do?
 - Creates FilterBar component so it can be used alone or on Table Toolbar.
 - Fixes vtex/styleguide/issues/530


react/components/EXPERIMENTAL_Conditions/Statement.js
```diff
@@ -68,36 +68,54 @@ class Statement extends React.Component {
       statements,
       statementIndex,
       labels,
+      ommitSubject,
```
jgfidelis: `ommit` is a type, if you want to say `omit` as "omitir", remove the extra m.

react/components/EXPERIMENTAL_FilterBar/FilterTag.js
```diff
@@ -0,0 +1,282 @@
+/* eslint-disable */
```
thiagomurakami: Did you forget to lint this file?

react/components/EXPERIMENTAL_Conditions/Statement.js
```diff
@@ -68,13 +68,15 @@ class Statement extends React.Component {
...
       labels,
+      onChangeObjectCallback,
     } = this.props
     const condition = statements[statementIndex]
     const atomProps = {
       statements: statements,
       options: options,
       isFullWidth: isFullWidth,
       statementIndex: statementIndex,
+      onChangeObjectCallback: onChangeObjectCallback,
```
thiagomurakami: You could declare it without the `: onChangeObjectCallback` if the key has the same name as the variable.

react/components/EXPERIMENTAL_Conditions/Atoms/ObjectAtom.js
```diff
@@ -11,6 +11,11 @@ class ObjectAtom extends React.Component {
     </div>
   )
 
+  handleOnChangeObjectCallback = (newValue, structure) => {
```
thiagomurakami: Do you need this function? I think you can pass `this.props. onChangeObjectCallback` directly.

react/components/EXPERIMENTAL_FilterBar/FilterTag.js
```diff
@@ -0,0 +1,282 @@
...
+
+const OPEN_MENU_STYLE = { backgroundColor: '#dbe9fd' }
+const emptyVirtualStatement = {
+  subject: null,
+  verb: null,
+  object: null,
+  error: null,
+}
+
+const filterStatementBySubject = (statements = [], sbj, options = {}) => {
```
thiagomurakami: Maybe using `subject` instead of `sbj` here would make it clearer.

react/components/EXPERIMENTAL_FilterBar/FilterTag.js
```diff
@@ -0,0 +1,282 @@
...
+      isMenuOpen: false,
+      virtualStatement: filterStatementBySubject(
+        props.statements,
+        props.subject,
+        props.options
+      )[0],
+    }
+  }
+
+  static getDerivedStateFromProps(nextProps, prevState) {
```
thiagomurakami: I think not returning anything when the if isn't met throws here... I'd double-check it, but maybe return `prevState` at the end would do the trick :)

react/components/EXPERIMENTAL_FilterBar/FilterTag.js
```diff
@@ -0,0 +1,282 @@
...
+      !this.filterMenuContainer.current.contains(e.target) &&
+      this.state.isMenuOpen
+    ) {
+      this.closeMenu()
+    }
+  }
+
+  componentWillUnmount() {
+    if (this.state.isMenuOpen) {
+      this.closeMenu()
```
thiagomurakami: Does calling a function that calls `setState` on `componentWillUnmount` trigger the error of trying to update an unmounted component?

react/components/EXPERIMENTAL_FilterBar/UncoupledStatement.js
```diff
@@ -0,0 +1,144 @@
...
+    return (
+      <div>
+        <div className="flex-column w-100 mv3">
+          <div
+            className={`flex w-100 items-start ${
+              isFullWidth ? 'flex-column items-stretch' : ''
+            }`}>
+            {omitSubject
+              ? omitVerbs
+                ? [<ObjectAtom key="object" {...atomProps} />]
```
thiagomurakami: This doesn't need to be an array

react/components/EXPERIMENTAL_FilterBar/UncoupledStatement.js
```diff
@@ -0,0 +1,144 @@
...
+                : [
```
thiagomurakami: ⛏ I prefer using `<Fragment>` instead of array in cases like this

react/components/EXPERIMENTAL_FilterBar/UncoupledStatement.js
```diff
@@ -0,0 +1,144 @@
...
+  options: PropTypes.object.isRequired,
+  /** Placeholder for subject dropdown */
+  subjectPlaceholder: PropTypes.string.isRequired,
+  /** Stretch component to 100% of the width */
+  isFullWidth: PropTypes.bool,
+  /** Statement change callback */
+  onChangeStatement: PropTypes.func,
+  /** To which row does this Statement belong to?  */
+  statementIndex: PropTypes.number,
+  /** Please use the 3 following ones with caution, i did not test them, so they can break everything */
```
thiagomurakami: Is it hard to test these?

react/components/EXPERIMENTAL_FilterBar/index.js
```diff
@@ -0,0 +1,235 @@
+/* eslint-disable camelcase */
```
thiagomurakami: ⛏ Naming variables without `_` is better than disabling this rule

react/components/EXPERIMENTAL_FilterBar/index.js
```diff
@@ -0,0 +1,235 @@
...
+  const newOptions = { ...options }
+  const optionsKeys = Object.keys(options)
+  optionsKeys.forEach(key => {
+    if (alwaysVisibleFilters.includes(key)) {
+      delete newOptions[key]
+    }
+  })
+  statements.forEach(st => {
+    if (st && st.object) {
+      delete newOptions[st.subject]
```
thiagomurakami: :drake-no:

react/components/EXPERIMENTAL_FilterBar/UncoupledStatement.js
ohmyguigs: yes 😞. I would have to develop this feature in the Conditions component to see if the prop fits or break it. Instead I made another component, so I guess there is no danger anymore. I'll remove the comment

react/components/EXPERIMENTAL_FilterBar/index.js
jgfidelis: You can just do a map taking out this key or using `omit` from ramda

react/components/EXPERIMENTAL_FilterBar/index.js
ohmyguigs: It's a pattern we chose for experimental components :(

react/components/EXPERIMENTAL_FilterBar/FilterTag.js
jgfidelis: Por baixo dos panos é o que o react faz se não me engano:
```
if (partialState === null || partialState === undefined) {
        // Null and undefined are treated as no-ops.
        return prevState;
      }
```

source: https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactUpdateQueue.js

react/components/EXPERIMENTAL_FilterBar/FilterTag.js
```diff
@@ -0,0 +1,282 @@
...
+      nextProps.subject &&
+      nextProps.statements.some(st => st.subject === nextProps.subject)
+    ) {
+      const statement = filterStatementBySubject(
+        nextProps.statements,
+        nextProps.subject,
+        nextProps.options
+      )[0]
+      return {
+        isMenuOpen: prevState.isMenuOpen,
```
jgfidelis: you dont need to pass the prevState variables, you just need to pass what changed.

react/components/EXPERIMENTAL_FilterBar/FilterTag.js
thiagomurakami: Ok. It's because I've had this error thrown on a component I made `Warning: Component.getDerivedStateFromProps():A valid state object (or null) must be returned. You have returned undefined.` 😓 

But if you're sure that React internals are accepting this now, no problem 👍

react/components/EXPERIMENTAL_FilterBar/index.js
thiagomurakami: Ohh, I thought you used this rule because of the `_st` variables :P

react/components/EXPERIMENTAL_FilterBar/UncoupledStatement.js
ohmyguigs: just to give context, before there was a prop `isRTL` that it runned a .reverse() in this array to adapt to eastern languages where speech structure is Object, Verb, Subject instead of Subject, Verb, Object.

Since I removed it, the array is not needed any more. Forgot it there.

cmdalbem:
Documentation is ready and available here: https://docs.google.com/document/d/18pJD3uJ1jz_AwyCsM6rMohg1uXICG6eZvx2accrEqHU/edit?usp=sharing

react/components/EXPERIMENTAL_FilterBar/UncoupledStatement.js
thiagomurakami: Oh, nice. Ok 👌

PR #542, opened 2019-02-15
Add Low prominence variation to tags

rsimoens:
Add a new prop to tags. Now a tag can have a low prominence version, with borders and no background color.

Our low prominence tags examples on styleguide weren't actually a prop, and the visual suggested wasn't attending minimum contrast standards.

##### Before
image
##### After
image

#### Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

### DIFF
--- a/react/components/Tag/index.js
+++ b/react/components/Tag/index.js
@@ -22,30 +22,49 @@ class Tag extends PureComponent {
   }
 
   render() {
-    const { children, onClick, disabled, color, bgColor } = this.props
+    const {
+      children,
+      onClick,
+      disabled,
+      color,
+      bgColor,
+      type,
+      variation,
+    } = this.props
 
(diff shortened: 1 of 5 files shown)


react/components/Tag/index.js
```diff
@@ -22,30 +22,49 @@ class Tag extends PureComponent {
...
+      variation,
+    } = this.props
 
     const baseClasses = 'br-pill t-small pv2 ph4 dib fw5'
 
     let theme = ''
-    switch (this.props.type) {
+    const variationIsLow = variation === 'low'
+
+    switch (type) {
```
JuanAlmeida12: You could replace this for a get function here and call it directly in `className`, What you think?

rsimoens: Could you show me another component with an example of that?
What are the benefits?

PR #547, opened 2019-02-18
Fix npm build

klzns:
Users of this library as a npm module are having issues to use it, since scripts inside the directory `modules` are not being published to npm.

See:
https://unpkg.com/@vtex/styleguide@8.18.3/

You can't find the module `withForwardedRef.js` there.


klzns:
cc @brunaaleixo

PR #555, opened 2019-02-26
Add the undismissable (by the user) Toast and use HOC instead of ref

estacioneto:
#### It will be needed to implement the "Offline Toast" feature into the store, for example.

image

image


react/components/ToastProvider/Toast.js
```diff
@@ -195,7 +199,7 @@ export default class Toast extends Component {
               </div>
             )}
           </div>
-          {onClose && (
+          {dismissable && (
```
klzns: I'm not sure here, but shouldn't it be `dismissable && onClose &&`?

estacioneto: I thought that too... The thing is the `onClose` prop is required and always passed by the `ToastManager`. The last one is because the manager must know when a toast is closed to show another toast if queued. So... I thought it wouldn't be a big deal to take it off.

klzns: Oh, I didn't see ToastManager always passed this function so.. 🤷‍♂️

PR #563, opened 2019-02-28
Add totalizers

kevinch:
Live: http://10.1.12.74:6060/#/Components/Display/Table

screen shot 2019-02-28 at 13 43 32

### DIFF
--- a/react/components/Table/Totalizers.js
+++ b/react/components/Table/Totalizers.js
@@ -5,7 +5,7 @@ import TotalizerLabel from './totalizers/TotalizerLabel'
 import TotalizerValue from './totalizers/TotalizerValue'
 import TotalizerIcon from './totalizers/TotalizerIcon'
 
-const BORDER_COLOR = 'b--light-gray'
+const BORDER_COLOR = 'b--muted-4'
 
 const TOTALIZERS_CLASSES = `w-100 flex flex-column flex-row-ns ba br3 mb5 ${BORDER_COLOR}`
 const TOTALIZER_CLASSES = `flex flex-column flex-auto pa4 ${BORDER_COLOR}`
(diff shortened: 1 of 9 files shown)


react/components/Table/Totalizers.js
```diff
@@ -0,0 +1,56 @@
+import React, { Fragment, PureComponent } from 'react'
+import PropTypes from 'prop-types'
+
+import TotalizerLabel from './totalizers/TotalizerLabel'
+import TotalizerValue from './totalizers/TotalizerValue'
+import TotalizerIcon from './totalizers/TotalizerIcon'
+
+const BORDER_COLOR = 'b--light-gray'
```
klzns: Use muted instead

react/components/Table/Totalizers.js
```diff
@@ -0,0 +1,56 @@
...
+
+const TOTALIZERS_CLASSES = `w-100 flex flex-column flex-row-ns ba br3 mb5 ${BORDER_COLOR}`
+const TOTALIZER_CLASSES = `flex flex-column flex-auto pa4 ${BORDER_COLOR}`
+
+class Totalizers extends PureComponent {
+  render() {
+    const { items } = this.props
+
+    if (items.length > 0) {
```
klzns: Invert this if and make it an early return

```js
if (item.length === 0) {
  return null
}
```

It should make it easier to read

react/components/Table/Totalizers.js
```diff
@@ -0,0 +1,56 @@
...
+      return (
+        <div id="totalizers" className={TOTALIZERS_CLASSES}>
```
klzns: If used twice in the page the id will clash. Why is it needed?

react/components/Table/Totalizers.js
```diff
@@ -0,0 +1,56 @@
...
+          {items.map((item, i) => {
+            const IS_NOT_LAST = items.length > 1 && i < items.length - 1
+            const EXTRA_BORDER = IS_NOT_LAST ? 'bb bb-0-ns br-ns' : ''
+
+            return (
+              <div className={`${TOTALIZER_CLASSES} ${EXTRA_BORDER}`} key={i}>
```
klzns: The key could be the item.label

react/components/Table/Totalizers.js
```diff
@@ -0,0 +1,56 @@
...
+        </div>
+      )
+    }
+
+    return null
+  }
+}
+
+Totalizers.propTypes = {
+  items: PropTypes.array.isRequired,
```
klzns: A shape here would be good

react/components/Table/index.js
```diff
@@ -248,6 +254,8 @@ Table.propTypes = {
     textOf: PropTypes.string,
     totalItems: PropTypes.number,
   }),
+  /** Totalizers property  */
+  totalizers: PropTypes.array,
```
klzns: A shape here would be good

react/components/Table/totalizers/TotalizerIcon.js
```diff
@@ -0,0 +1,36 @@
+import React, { PureComponent } from 'react'
+import PropTypes from 'prop-types'
+
+const CIRCLE_DIAMETER = '2.4rem'
+
+class TotalizerIcon extends PureComponent {
+  render() {
+    const {
+      item: { icon, iconBackgroundColor },
```
klzns: There's no iconBackgroundColor in shape of prop.

Also I think it's better as it is in the shape, limiting the option of colors to success and error

react/components/Table/totalizers/TotalizerLabel.js
```diff
@@ -0,0 +1,16 @@
+import React, { PureComponent } from 'react'
+import PropTypes from 'prop-types'
+
+class TotalizerLabel extends PureComponent {
```
klzns: I think this could be a function component

react/components/Table/totalizers/TotalizerValue.js
```diff
@@ -0,0 +1,36 @@
+import React, { PureComponent } from 'react'
+import PropTypes from 'prop-types'
+import config from 'vtex-tachyons/config.json'
+
+import Spinner from '../../Spinner'
+
+class TotalizerValue extends PureComponent {
```
klzns: I think this could be a function component

react/components/Table/totalizers/TotalizerValue.js
```diff
@@ -0,0 +1,36 @@
...
+    if (value) {
+      return <div className="f4 fw5 c-on-base">{value}</div>
+    }
+
+    return null
+  }
+}
+
+TotalizerValue.propTypes = {
+  item: PropTypes.object,
```
klzns: Define a shape

react/components/Table/totalizers/TotalizerValue.js
```diff
@@ -0,0 +1,36 @@
...
+    const {
+      item: { value, isLoading },
+    } = this.props
+
+    if (isLoading) {
+      return (
+        <div className="mt2">
+          <Spinner
+            size="14"
+            color={config.semanticColors.background['base--inverted']}
```
klzns: Use class to color the icon

react/components/Table/totalizers/TotalizerValue.js
```diff
@@ -0,0 +1,36 @@
...
+          />
+        </div>
+      )
+    }
+
+    if (value) {
```
klzns: Invert this, the last case should be the default case.

```js
if (value == null) {
  return null
}
```

Also it should test if it's null or undefined like above, because if the value is zero it would return null.

react/components/Table/Totalizers.js
kevinch: Good point and not useful.

react/components/Table/Totalizers.js
kevinch: No problem with spaces in the string?

react/components/Table/totalizers/TotalizerIcon.js
kevinch: I've discussed it with @rsimoens and we will let it "open".

react/components/Table/totalizers/TotalizerValue.js
kevinch: How so?

react/components/Table/Totalizers.js
klzns: Nope

react/components/Table/totalizers/TotalizerValue.js
klzns:
```js
<div className="mt2 c-muted-5">
  <Spinner size="14" color="currentColor">
</div>
```

PR #567, opened 2019-02-28
Table line actions

ericreis:

Add a column used for actions to be executed on each line. This is controled by a prop called `lineActions`:

```js
lineActions: PropTypes.arrayOf(
  PropTypes.shape({
    label: PropTypes.func,
    isDangerous: PropTypes.bool,
    onClick: PropTypes.func,
  })
),
```

1. Run `yarn start`
2. Navigate to http://localhost:6060/#/Components/Display/Table
3. The 2nd example is for this feature

styleguide table-line-actions

#### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

### DIFF
--- a/react/components/Table/SimpleTable.js
+++ b/react/components/Table/SimpleTable.js
@@ -42,7 +42,6 @@ class SimpleTable extends Component {
   handleRowHover = rowIndex => {
     const { onRowClick } = this.props
     const { isLineActionsHovered } = this.state
-    console.log(isLineActionsHovered)
     if (onRowClick && !isLineActionsHovered) {
       this.setState({
         hoverRowIndex: rowIndex,
@@ -271,9 +270,9 @@ class SimpleTable extends Component {
                           onClick={
                             onRowClick &&
                             property !== '_VTEX_Table_Internal_lineActions'
-                              ? e =>
+                              ? event =>
                                   onRowClick({
-                                    e,
+                                    event,
                                     index: rowIndex,
                                     rowData,
                                   })
--- a/react/components/Table/SimpleTable.js
+++ b/react/components/Table/SimpleTable.js
@@ -72,11 +72,9 @@ class SimpleTable extends Component {
         cellRenderer: ({ rowData }) => {
           return (
             <ActionMenu
-              icon={<OptionsDots />}
-              hideCaretIcon
               buttonProps={{
                 variation: 'tertiary',
-                icon: true,
+                icon: <OptionsDots />,
               }}
               options={lineActions.map(action => ({
                 ...action,
@@ -251,11 +249,10 @@ class SimpleTable extends Component {
                               fullWidthColWidth
                             ),
                           }}
-                          className={`flex items-center w-100 h-100 ph4 bb b--muted-4 ${
-                            property === '_VTEX_Table_Internal_lineActions'
-                              ? ''
-                              : 'truncate'
-                          } ${disableHeader && rowIndex === 0 ? 'bt' : ''} ${
+                          className={`flex items-center w-100 h-100 ph4 bb 
+                            b--muted-4 truncate ${
+                              disableHeader && rowIndex === 0 ? 'bt' : ''
+                            } ${
                             onRowClick && rowIndex === hoverRowIndex
                               ? 'pointer bg-near-white c-link'
                               : ''
--- a/react/components/Table/SimpleTable.js
+++ b/react/components/Table/SimpleTable.js
@@ -11,6 +11,8 @@ import ActionMenu from '../ActionMenu'
 const ARROW_SIZE = 11
 const HEADER_HEIGHT = 36
 const DEFAULT_COLUMN_WIDTH = 200
+const LINE_ACTIONS_COLUMN_WIDTH = 70
+const NO_TITLE_COLUMN = ' '
 
 class SimpleTable extends Component {
   constructor(props) {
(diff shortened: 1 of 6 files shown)


react/components/Table/SimpleTable.js
```diff
@@ -72,9 +102,13 @@ class SimpleTable extends Component {
       updateTableKey,
       rowHeight,
       fullWidth,
+      lineActions,
       loading,
     } = this.props
     const { hoverRowIndex } = this.state
+
+    if (lineActions)
+      schema.properties = this.addLineActionsToSchema(schema, lineActions)
```
klzns: Not sure if this mutation might cause an issue 🤔

react/components/Table/SimpleTable.js
```diff
@@ -58,6 +61,33 @@ class SimpleTable extends Component {
...
   }
 
+  addLineActionsToSchema(schema, lineActions) {
+    return (schema.properties = {
+      ...schema.properties,
+      // eslint-disable-next-line camelcase
+      _VTEX_Table_Internal_lineActions: {
+        type: 'any',
+        title: ' ',
+        width: 70,
```
ohmyguigs: ⛏this could be a const outside the class like the other static numbers. 
LINE_ACTIONS_COLUMN_WIDTH or something like that.

react/components/Table/SimpleTable.js
```diff
@@ -58,6 +61,33 @@ class SimpleTable extends Component {
       : DEFAULT_COLUMN_WIDTH
   }
 
+  addLineActionsToSchema(schema, lineActions) {
+    return (schema.properties = {
```
ohmyguigs: this is a little weird. You could return a new schema since you receive the schema as param.
```js
return {
  properties: {
     ...schema.properties,
     .......here your code like it is...
  }
}
```

react/components/Table/SimpleTable.js
ericreis: This would only add one column with the prop `lineActions` is passed. What do you think it could cause?

react/components/Table/SimpleTable.js
```diff
@@ -36,10 +41,16 @@ class SimpleTable extends Component {
 
   handleRowHover = rowIndex => {
     const { onRowClick } = this.props
-    if (onRowClick) {
+    const { isLineActionsHovered } = this.state
+    console.log(isLineActionsHovered)
```
lbebber: :eyes:

react/components/Table/SimpleTable.js
```diff
@@ -3,11 +3,16 @@ import PropTypes from 'prop-types'
...
+import OptionsDots from '../icon/OptionsDots'
 import EmptyState from '../EmptyState'
 import Spinner from '../Spinner'
+import ActionMenu from '../ActionMenu'
+
 const ARROW_SIZE = 11
 const HEADER_HEIGHT = 36
 const DEFAULT_COLUMN_WIDTH = 200
+const LINE_ACTIONS_COLUMN_WIDTH = 70
+const NO_TITLE_COLUMN = ' '
```
lbebber: not sure why this needs to be an external variable, but ok

react/components/Table/SimpleTable.js
```diff
@@ -215,16 +260,23 @@ class SimpleTable extends Component {
...
                             onRowClick && rowIndex === hoverRowIndex
                               ? 'pointer bg-near-white c-link'
                               : ''
                           } ${columnIndex === 0 && fixFirstColumn ? 'br' : ''}`}
                           onClick={
-                            onRowClick
-                              ? e => onRowClick({ e, index: rowIndex, rowData })
+                            onRowClick &&
+                            property !== '_VTEX_Table_Internal_lineActions'
+                              ? e =>
```
lbebber: ⛏ `event` would be nicer, but its your call

react/components/Table/SimpleTable.js
lbebber: or why this is a whitespace instead of `''`

react/components/Table/SimpleTable.js
ericreis: @lbebber I'm not sure I understand what you are suggesting 🤔

react/components/Table/SimpleTable.js
ericreis: already removed it 😄

react/components/Table/SimpleTable.js
ohmyguigs: seems ok now :)

react/components/Table/SimpleTable.js
lbebber: renaming from `e` to `event`, to make it clearer. just a suggestion though

react/components/Table/SimpleTable.js
ohmyguigs: @lbebber when rendering the column, table checks for the title with something like `if (title)` so  `''` would be considered null

react/components/Table/SimpleTable.js
ohmyguigs: @ericreis swap `e` for `event` for easier understanding in the callback object param.

react/components/Table/SimpleTable.js
ericreis: @lbebber I don't like this behaviour too, I'll check with @ohmyguigs if it makes sense to change this 😃

PR #572, opened 2019-03-08
fix withForwardedRef proptypes

jgfidelis:
`oneOfType` should receive an array


ericreis:
Duplicate of #573. Since it's best if we don't add this kind of technical fixes in the CHANGELOG, maybe should keep #573. @jgfidelis @klzns

PR #581, opened 2019-03-15
Export ToastContext to allow useContext and ToastManager is now a HOC

estacioneto:
**There is no breaking changes. ToastProvider API is the same.**

As the title says, with the `ToastContext` exported, we can now use it with Hooks (for those who don't want to use the `withToast` HOC in functional components):

```jsx
import { useContext } from 'react'
import { ToastContext } from 'vtex.styleguide'

function Component() {
  ...
  const { showToast, hideToast, toastState } = useContext(ToastContext)
  ...
}
```

### How to test it

[Access the workspace](http://toast--storecomponents.myvtex.com) and open the dev tools and click the `Network > Offline` option (or turn off your device's network). Then click the buy button to see the offline toast be replaced but when the buy button toast goes away, the offline toast goes up again.

### DIFF
--- a/react/ToastContext.js
+++ b/react/ToastContext.js
@@ -1,0 +1,3 @@
+import { ToastContext } from './components/ToastProvider/index'
+
+export default ToastContext
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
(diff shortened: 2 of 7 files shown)


react/ToastContext.js
```diff
@@ -0,0 +1,3 @@
+import { ToastContext } from './components/ToastProvider/index'
+
+export default ToastContext
```
klzns: Just check if this breaks while linking the app

estacioneto: Done :grin:

PR #590, opened 2019-03-25
Add 'align' and 'direction' props to DatePicker

alcar:
**Right-aligned:**
image

<hr>

**Upwards:**
image

### DIFF
--- a/react/components/DatePicker/index.js
+++ b/react/components/DatePicker/index.js
@@ -23,9 +23,9 @@ class DatePicker extends Component {
     const { align, direction, size, useTime } = this.props
 
     const isRightAligned = align === 'right'
-    const isUpwardsDirected = direction === 'up'
+    const isUpwards = direction === 'up'
 
-    if (!isRightAligned && !isUpwardsDirected) {
+    if (!isRightAligned && !isUpwards) {
       return undefined
     }
 
@@ -39,13 +39,11 @@ class DatePicker extends Component {
       small: -432,
     }
 
-    const offsetY = isUpwardsDirected
-      ? offsetYBySize[size] - (useTime ? 18 : 0)
-      : 0
+    const offsetY = isUpwards ? offsetYBySize[size] - (useTime ? 18 : 0) : 0
 
     return {
       flip: {
-        enabled: !isUpwardsDirected,
+        enabled: !isUpwards,
       },
       keepTogether: {
         enabled: false,
--- a/react/components/DatePicker/index.js
+++ b/react/components/DatePicker/index.js
@@ -20,13 +20,16 @@ class DatePicker extends Component {
   }
 
   get popperModifiers() {
-    const { flipped, rightAligned, size, useTime } = this.props
+    const { align, direction, size, useTime } = this.props
 
-    if (!flipped && !rightAligned) {
+    const isRightAligned = align === 'right'
+    const isUpwardsDirected = direction === 'up'
+
+    if (!isRightAligned && !isUpwardsDirected) {
       return undefined
     }
 
-    const offsetX = rightAligned
+    const offsetX = isRightAligned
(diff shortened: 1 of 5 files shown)


react/components/DatePicker/index.js
```diff
@@ -153,6 +191,8 @@ DatePicker.propTypes = {
   readOnly: PropTypes.bool,
   /** Spec attribute  */
   required: PropTypes.bool,
+  /** Whether or not the popper should be aligned along the right side of the input */
+  rightAligned: PropTypes.bool,
```
lbebber: this prop could follow the standard set by `ActionMenu`, i.e. `align: left|right`

react/components/DatePicker/index.js
```diff
@@ -123,6 +159,8 @@ DatePicker.propTypes = {
   excludeDates: PropTypes.arrayOf(PropTypes.instanceOf(Date)),
   /** Times to be excluded  */
   excludeTimes: PropTypes.arrayOf(PropTypes.instanceOf(Date)),
+  /** Whether or not the popper should be flipped vertically */
+  flipped: PropTypes.bool,
```
lbebber: this could be something like, `direction: up|down`, or `verticalAlignment: top|bottom`, or something of the sort

(when there is no "right" value, or there is no clear semantic of "on/off", it's generally better to use an enum rather than a boolean)

ericreis: Also, when I read `flipped` I imagine the **DatePicker** upside down, haha

react/components/DatePicker/index.js
```diff
@@ -19,6 +19,43 @@ class DatePicker extends Component {
     this.handleLocaleChange(props.locale)
   }
 
+  get popperModifiers() {
+    const { align, direction, size, useTime } = this.props
+
+    const isRightAligned = align === 'right'
+    const isUpwardsDirected = direction === 'up'
```
lbebber: `isUpwards` is enough (the suffix "ward"/"wards" already implies direction)

react/components/DatePicker/index.js
```diff
@@ -19,6 +19,43 @@ class DatePicker extends Component {
...
+
+    if (!isRightAligned && !isUpwardsDirected) {
+      return undefined
+    }
+
+    const offsetX = isRightAligned
+      ? (size === 'large' ? -91 : -136) - (useTime ? 130 : 0)
```
lbebber: where do these number come from?

react/components/DatePicker/index.js
```diff
@@ -19,6 +19,43 @@ class DatePicker extends Component {
...
+      : 0
+
+    const offsetYBySize = {
+      large: -448,
```
lbebber: these too. are they always fixed regardless of CSS?

react/components/DatePicker/index.js
alcar: It can also be used for movement, e.g. "moving upwards" or "rising upwards". I don't mind renaming it though, haha.

react/components/DatePicker/index.js
alcar: Well, the `size` part comes from pure observation, while the `useTime` one comes from the time column width, that is 130px wide.

react/components/DatePicker/index.js
alcar: Also observation. I'd say yes, because all that matters is the size of the input. Can you think of an edge case that I might've missed? Perhaps size transformations?

react/components/DatePicker/index.js
lbebber: so does `Directed`

ericreis: I personally prefer `direction`, but I do not oppose to renaming it 😄

PR #594, opened 2019-04-02
Add TimePicker component

### LINKED ISSUE #65: Timepicker
Substituto para o `<input type="time" />`

ericreis:
This component uses the **DatePicker** component as the base. That way every new feature/fix we implement in the **DatePicker** we will also be applied to the **TimePicker** component.

Closes #65.

### DIFF
--- a/react/components/TimePicker/README.md
+++ b/react/components/TimePicker/README.md
@@ -1,0 +1,99 @@
+#### A TimePicker lets the user select a single time.
+
+### 👍 Dos
+
+### 👎 Don'ts
+
+### Supported locales
+
+You can check out the list of supported locales [here](https://github.com/date-fns/date-fns/blob/master/src/locale/index.js).
+
+Simple _TimePicker_
+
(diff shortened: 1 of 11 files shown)


ericreis:
Still missing Do's and Dont's and better documentation examples.

@cmdalbem Can you help me with this?

react/components/TimePicker/README.md
```diff
@@ -0,0 +1,99 @@
+#### A TimePicker lets the user select a single time.
+
+### 👎 Don'ts
+
+- Time suggestions should help the user to find an option more easily given its context of use (per hour, every half hour, 15 minutes, ...) so we recommend not using intervals shorter than 5 minutes.
```
kevinch:
```suggestion
- Time suggestions should help the user to find an option more easily given its context of use (per hour, every half hour, 15 minutes, etc.) so we do not recommend using intervals shorter than 5 minutes.
```

PR #603, opened 2019-04-05
dont pass null to oneOfType function

jgfidelis:
FIx error:
Screen Shot 2019-04-05 at 12 11 27 PM

wks: https://teste--storecomponents.myvtex.com/

There is no need to pass oneOfType([null, mytype]), this is already implicit if you don't insert a `.isRequired` after your type.


klzns:
Please do this to the master branch

PR #604, opened 2019-04-05
Feature - Date and time range props

alcar:
image


alcar:
> I think the range props should be ignored from the documentation for now. In the near future we should create another component for the range logic and use those props only internally.

I don't think that's a good idea. Styleguide is an open-source project, so not documenting a feature does not mean it won't be used. Also, if we end up creating a separate component for time and date ranges, and then decide to remove the props I've just added, it's still a breaking change and should be released as a new major version, whether it's documented or not.

> Also, I think the changelog could be more self explanatory.

I agree that I could've been more descriptive; I'll rewrite it right away!

Issue #479, opened 2019-01-03
Table - Line and bulk actions


cmdalbem:
Here is a proposal of a new feature for the Table component that allows us to have actions per line and bulk actions to be executed on multiple lines.
The developer has to be able to define a list of actions, and for each of them if they can be executed on multiple lines. For a list of actions it's possible to highlight one main action.
It's important to note there are two possible states for the "all selected" case. In one case you're selecting all the visible options, in the other your selecting in fact ALL the available data, including other pages. This solutions was inspired by how Gmail deals with the same problem.

![bulk actions](https://user-images.githubusercontent.com/467471/50651420-65c6fb00-0f6a-11e9-84fa-96a4144ad9a3.png)

cmdalbem:
Closed by https://github.com/vtex/styleguide/pull/612

PR #612, opened 2019-04-11
Feature/table bulk actions

kevinch:
Figma:
https://www.figma.com/file/aWaWGvrDNlyNsQKYfrwtsG/Tables?node-id=10041%3A14297

Missing features / bugs:
- [x] animation on opening
- [x] bug line height checkbox icon
- [x] checkbox semi selected state
- [x] behavior with onRowClick (prevent row click)
- [x] bug line height x selected items
- [x] selected line style
- [x] write doc
- [x] selected rows does not go to 0 while the closing animation happens 
- [x] fix blue border on primary buttons

### DIFF
--- a/react/components/Table/BulkActions.js
+++ b/react/components/Table/BulkActions.js
@@ -22,7 +22,9 @@ class BulkActions extends PureComponent {
     } = this.props
 
     const hasBulkActions = hasPrimaryBulkAction || hasSecondaryBulkActions
-    const hasRowsSelected = selectedRows.length > 0
+    const selectedRowsLength = selectedRows.length
+    const hasRowsSelected = selectedRowsLength > 0
+    const hasOneRowSelected = selectedRowsLength === 1
 
     const bulkActionsReturnedParameters = allLinesSelected
       ? { allLinesSelected: true }
@@ -70,7 +72,10 @@ class BulkActions extends PureComponent {
         <div className="tr flex flex-row items-center">
           {!allLinesSelected && bulkActions && bulkActions.texts && (
             <span className="mr4 c-muted-4">
-              {selectedRows.length} {bulkActions.texts.rowsSelected}
+              {selectedRowsLength}{' '}
+              {hasOneRowSelected
+                ? bulkActions.texts.rowSelected
+                : bulkActions.texts.rowsSelected}
             </span>
           )}
           <span className="mr2">
--- a/react/components/Table/BulkActions.js
+++ b/react/components/Table/BulkActions.js
@@ -53,7 +53,9 @@ class BulkActions extends PureComponent {
                   variation="secondary"
                   size="small"
                   onClick={() =>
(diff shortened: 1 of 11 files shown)


kevinch:
Closes https://github.com/vtex/styleguide/issues/479

kevinch:
wip at https://github.com/vtex/vtex-tachyons/pull/29

react/components/Table/CheckboxContainer.js
```diff
@@ -0,0 +1,39 @@
+import React, { PureComponent } from 'react'
+import PropTypes from 'prop-types'
+
+import Checkbox from '../Checkbox'
+
+class CheckboxContainer extends PureComponent {
```
klzns: Use Component instead since the component has a usage like:

```jsx
onClick={() => doSometing(id)}
```
This arrow function will always make the component rerender making the checks by PureComponent more expensive than using Component.

react/components/Table/BulkActions.js
```diff
@@ -0,0 +1,136 @@
...
+            )}
+          </div>
+        )}
+        <div className="tr flex flex-row items-center">
+          {!allLinesSelected && bulkActions && bulkActions.texts && (
+            <span className="mr4 c-muted-4">
+              <BulkActionsSelectedRows
+                selectedRowsLength={selectedRowsLength}
+              />{' '}
+              {hasOneRowSelected
```
klzns: You can't assume that the string will be formated as:

`NUMBER rows selected`

Maybe there's a language that must render something like:

`Rows selected NUMBER`

Maybe it's better to delegate to the client calling a callback:

Usage by the client:
```
selectedRows={(number) =>
  intl.formatMessage({ id: 'selectedRows', values={{number}})  
)
```

react/components/Table/BulkActions.js
```diff
@@ -0,0 +1,136 @@
...
+                : bulkActions.texts.rowsSelected}
+            </span>
+          )}
+          <span className="mr2">
+            {allLinesSelected ? (
+              bulkActions &&
+              bulkActions.texts && (
+                <span>
+                  {bulkActions.texts.allRowsSelected}{' '}
+                  <span className="b">{bulkActions.totalItems}</span>
```
klzns: Same thing here

react/components/Table/BulkActions.js
kevinch: Good point.

react/components/Table/CheckboxContainer.js
kevinch: Updated.

react/components/Table/BulkActions.js
kevinch: Updated.

react/components/Table/BulkActions.js
kevinch: Updated.

PR #614, opened 2019-04-11
Feature/stable filter bar

ohmyguigs:
# What does this PR do?

 -  fixes vtex/styleguide/issues/606
 - fixes vtex/styleguide/issues/607
 - Takes FilterBar out of EXPERIMENTAL section


react/components/FilterBar/Atoms/SubjectAtom.js
```diff
@@ -54,14 +54,23 @@ class SubjectAtom extends React.Component {
...
       <div className={`mh3 ${isFullWidth ? 'pb3' : ''}`}>
-        <Dropdown
+        <Select
           ref={forwardedRef}
           placeholder={placeholder}
           options={uniqueOptions}
-          value={!condition.subject ? '' : condition.subject || ''}
-          onChange={(e, value) => {
-            this.handleChangeStatement(value, 'subject')
+          defaultValue={
```
ericreis: I don't usually use defaultValue, I use the value prop. Generally the select value is passed as a prop and should be used as the prop value in my opinion, but we can discuss your use case.

PR #616, opened 2019-04-12
Allow disabling new line action button in table toolbar

ericreis:


react/components/Table/Toolbar.js
```diff
@@ -370,6 +370,11 @@ Toolbar.propTypes = {
       alignMenu: PropTypes.oneOf(['right', 'left']),
       isLoading: PropTypes.bool,
     }),
+    newLine: PropTypes.shape({
+      label: PropTypes.string,
```
kevinch: Sounds to me that this is one should be required.

react/components/Table/Toolbar.js
```diff
@@ -370,6 +370,11 @@ Toolbar.propTypes = {
...
+      handleCallback: PropTypes.func,
```
kevinch: Same here.

ohmyguigs: this ones can't be required since newLine button is optional. The use of this props is what makes the button appear.

react/components/Table/index.js
```diff
@@ -272,8 +272,8 @@ Table.propTypes = {
       alignMenu: PropTypes.oneOf(['right', 'left']),
     }),
     newLine: PropTypes.shape({
-      label: PropTypes.string,
-      handleCallback: PropTypes.func,
+      label: PropTypes.string.isRequired,
+      handleCallback: PropTypes.func.isRequired,
     }),
```
kevinch: Didn't you forget the `disabled` prop here?

PR #618, opened 2019-04-16
Table's toolbar without the search bar is aligned to the left

thiagomurakami:
This PR fixes the case where you want to use the `Toolbar` without the search bar.

Before: 
Screen Shot 2019-04-16 at 17 05 55

After:
Screen Shot 2019-04-16 at 17 05 28


react/components/Table/Toolbar.js
```diff
@@ -120,7 +121,11 @@ class Toolbar extends PureComponent {
       density.highOptionLabel
```
kevinch: Missing changelog.

PR #636, opened 2019-05-07
Remove focus outline from the Collapsible component

Lurian:
Remove focus outline for the `Collapsible` component for non-keyboard users.

This PR adds a [pollyfil](https://github.com/WICG/focus-visible) for the [`focus-visible`](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible) class.

Screen Shot 2019-05-07 at 3 38 28 PM

The above ring will not show up when using a device for navigation different from a keyboard (e.g. using a mouse)s


react/components/Collapsible/index.js
```diff
@@ -72,7 +73,7 @@ class Collapsible extends Component {
     const color = muted ? 'c-muted-3' : 'c-action-primary'
 
     return (
-      <div>
+      <div className={`${jsFocusVisible}`}>
```
jgfidelis: you can just do `className={jsFocusVisible}`

PR #642, opened 2019-05-09
Fix table resize

salesfelipe:

- Fix table resize when using fullwidth and the `updateTableKey` prop.

#### Types of changes

### DIFF
--- a/react/components/Table/SimpleTable.js
+++ b/react/components/Table/SimpleTable.js
@@ -134,11 +134,8 @@ class SimpleTable extends Component {
           </div>
         ) : (
           <div>
-            <AutoSizer>
+            <AutoSizer key={updateTableKey}>
               {({ width }) => {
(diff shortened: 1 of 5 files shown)


react/components/Table/SimpleTable.js
```diff
@@ -132,11 +132,8 @@ class SimpleTable extends Component {
           </div>
         ) : (
           <div>
-            <AutoSizer>
+            <AutoSizer key={updateTableKey}>
```
ohmyguigs: maybe we should guarantee this is always a string. I dunno the behaviour react makes if a key is undefined or null
```suggestion
            <AutoSizer key={`vtex-table__${updateTableKey}`}>
```

salesfelipe: I've just tested it, it works like if you were not passing at all.

ohmyguigs: LGTM then

Issue #647, opened 2019-05-13
Table shows two scrollbars instead of only one


luciannojunior:
Using **Windows** or **Linux**, the _Table_ component displays two scrollbars when should only display one (the horizontal one), as it can be seen below:
![image](https://user-images.githubusercontent.com/18706156/57628334-1c1d9b00-7570-11e9-9f8d-64542601767d.png)
_[Table](https://styleguide.vtex.com/#/Components/Display/Table) on Chrome for Windows_

This problems **doesn't happen on MacOS**, where only the x-axis scrollbar is seen when scrolling:
![image](https://user-images.githubusercontent.com/18706156/57628462-5f780980-7570-11e9-94bd-b938f126e6cf.png)
_[Table](https://styleguide.vtex.com/#/Components/Display/Table) on Chrome for MacOS_

X-axis' scrollbar height is probably the cause for this issue.

PR #649, opened 2019-05-14
add height offset to horizontal scrollbar in Table component

viniagostini:
as the title says

it resolves this issue: https://github.com/vtex/styleguide/issues/647

prints from before the fix are in the issue  

**After the fix:**
chrome:
fixed_chrome

firefox:
fixed_firefox

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### Fixed
-- Add offset in Table height so the horizontal scroll bar do not overlap its content.
+- Add offset in Table height so the horizontal scroll bar does not overlap its content.
 
 ## [8.46.1] - 2019-05-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
 ### Fixed
-- Add offset in Table height so the horizontal scroll bar does not overlap its content.
+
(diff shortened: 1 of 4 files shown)


react/components/Table/index.js
```diff
@@ -17,6 +17,7 @@ import { pluck } from 'ramda'
 
 const TABLE_HEADER_HEIGHT = 36
 const EMPTY_STATE_SIZE_IN_ROWS = 5
+const X_SCROLLBAR_OFFSET = 17
```
lbebber: the scrollbar size varies depending on the operating system (and user settings on top of that—some accessibility settings make the scrollbar large, for example).

can you find a way for this not to be hardcoded?

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- Add offset in Table height so the horizontal scroll bar do not overlap its content.
```
kevinch:
```suggestion
- Add offset in Table height so the horizontal scroll bar does not overlap its content.
```

react/components/Table/index.js
viniagostini: yeah @lbebber i found some solutions, i think this is the cleanest one:
``` javascript
const getScrollbarWidth = () => window.innerWidth - document.documentElement.clientWidth;
```
what do you think?

i even tested here changing my scrollbar width on the browser :smile:

lbebber: @viniagostini LGTM but:
- check the existence of `window`, `document`, and `document.documentElement`, in case it runs on SSR
- check if it works across multiple browsers (I think it should, but just in case)
- make sure it always returns a number (and not `NaN` or something)

viniagostini: Done! :smiley:

PR #536, opened 2019-02-13
Fix/spinner animation

lbebber:
Fixes spinner animations not working while the page is loading

PR #662, opened 2019-05-24
Misc improvements

thiagomurakami:
1. Fix `Button`'s disabled border.
2. Set message shown to user on `Select` when the `creatable` prop is `true`.
3. Control `resize` behavior of `Textarea`.
4. Get commits from [https://github.com/vtex/styleguide/pull/536/](https://github.com/vtex/styleguide/pull/536/) that weren't merged.

Old disabled Button:
Screen Shot 2019-05-24 at 15 35 17

New disabled Button:
Screen Shot 2019-05-24 at 15 35 05

You can now translate the message shown when creating an option using `Select`:
Screen Shot 2019-05-24 at 15 36 02

Old resize behavior allows the user to break the UI:
Screen Shot 2019-05-24 at 15 40 02

Allows `vertical` resize only in `Textarea`:
ScreenRecording20190524at1

`Spinner` using CSS animations doesn't stop spinning when main thread is busy processing stuff :P

#### Types of changes

### DIFF
--- a/react/components/EXPERIMENTAL_Select/index.js
+++ b/react/components/EXPERIMENTAL_Select/index.js
@@ -49,6 +49,7 @@ class Select extends Component {
       defaultValue,
       disabled,
       errorMessage,
+      formatCreateLabel,
       label,
       loading,
       multi,
@@ -88,6 +89,7 @@ class Select extends Component {
         Placeholder,
       },
       defaultValue,
+      formatCreateLabel,
       getOptionValue,
       isClearable: clearable,
       isDisabled: disabled,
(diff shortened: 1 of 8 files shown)


react/components/EXPERIMENTAL_Select/index.js
```diff
@@ -212,6 +214,8 @@ Select.propTypes = {
   disabled: PropTypes.bool,
   /** Error message, e.g., validation error message. */
   errorMessage: PropTypes.string,
+  /** Gets the label for the "Create new..." option in the menu. ({inputValue}) => string | null */
+  formatCreateLabel: PropTypes.func,
```
lbebber: how about just `createLabel`?

PR #671, opened 2019-06-06
Feature/dropzone

kevinch:
Add Dropzone component.

Screen Shot 2019-09-06 at 15 34 07

#### Types of changes


cmdalbem:
@kevinch I noticed you tagged this with `Designing`. Do you need help with that?

kevinch:
@cmdalbem i've used as base the component in the pickup-points admin. But if you're willing to give it an update please do so.

kevinch:
A design review would be good at this point. Component is ready imo.

kevinch:
# Change base to 9.x
It uses hooks only available in 9.x major.

kevinch:
Will be moved to 9.x.

kevinch:
@ohmyguigs no idea...

alinevillaca:
On Figma, there is a middle state between the default state and "drag hovering" state (the blue one) to indicate where the drag zone is. Is it possible to add?
image

The loaded state has a minor horizontal misalignment 
image

react/components/Dropzone/index.js
```diff
@@ -0,0 +1,136 @@
...
+import ReactDropZone from 'react-dropzone'
+
+import FileIcon from './FileIcon'
+import Spinner from '../Spinner'
+import ButtonWithIcon from '../ButtonWithIcon'
+import IconClose from '../icon/Close'
+
+const dropzoneRef = createRef()
+
+class Dropzone extends PureComponent {
```
matheusps: 💡Using hooks and FC would shorten up a lot of this code.

react/components/Dropzone/index.js
```diff
@@ -0,0 +1,136 @@
...
+    const initialState = !isHovering && !fileDropped
+
+    let dropzoneContainerClasses =
+      'flex flex-column items-center justify-center b--light-blue ba br2 bw1 b--dashed '
+    let iconHolderClasses =
+      'flex items-center justify-center pa6  bg-near-white '
+    const droppedContainerClasses =
+      'bg-light-silver flex flex-row justify-between w-100 items-top'
+
+    const holderSize = '120px'
```
matheusps: Maybe define it as a constant outside of the component, like `HOLDER_SIZE`.

react/Dropzone.js
```diff
@@ -0,0 +1,3 @@
+import Dropzone from './components/Dropzone/index'
```
matheusps: Could you do this in `typescript`? 🤔

kevinch:
@alinevillaca nope.

react/components/Dropzone/index.js
kevinch: Not too sure since it would be the only one used only once. A const file makes sense imo if used in multiple places or several times.

matheusps: Ok, makes sense.

react/Dropzone.js
kevinch: Will think about it... But only next week.

kevinch:
@ericreis Will think about it... But only next week.

PR #674, opened 2019-06-10
Fix/modal-scrolltop

joaopedroxavier:

This pull request is a proposal to fix an unexpected behavior in Modals.

Some Modals, when rendered, comes with a shadow. This shadow is rendered when the user scrolls the content from the Modal, so it shouldn't be rendered in this case. This is due to the comparison of the value of 'scrollTop' when it's still undefined. After some conversation with @samuraiexx, a given workaround for that was to always check if this state is incorrectly set.

#### Types of changes

### DIFF
--- a/react/yarn.lock
+++ b/react/yarn.lock
@@ -746,11 +746,6 @@ raf@^3.4.0:
   dependencies:
     performance-now "^2.1.0"
 
-ramda@^0.26.1:
-  version "0.26.1"
-  resolved "https://registry.yarnpkg.com/ramda/-/ramda-0.26.1.tgz#8d41351eb8111c55353617fc3bbffad8e4d35d06"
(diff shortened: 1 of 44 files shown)


react/yarn.lock
```diff
@@ -746,11 +746,6 @@ raf@^3.4.0:
...
     performance-now "^2.1.0"
 
-ramda@^0.26.1:
-  version "0.26.1"
-  resolved "https://registry.yarnpkg.com/ramda/-/ramda-0.26.1.tgz#8d41351eb8111c55353617fc3bbffad8e4d35d06"
-  integrity sha512-hLWjpy7EnsDBb0p+Z3B7rPi3GDeRG5ZtiI33kJhTt+ORCd38AbAIjB/9zRIUoeTbE/AVX5ZkU7m6bznsvrf8eQ==
-
 react-color@^2.17.0:
   version "2.17.0"
   resolved "https://registry.yarnpkg.com/react-color/-/react-color-2.17.0.tgz#e14b8a11f4e89163f65a34c8b43faf93f7f02aaa"
```
kevinch: Missing changelog entry.

joaopedroxavier: Added! 😄

Issue #688, opened 2019-06-13
BulkActions - Select All seems reversible


rafarubim:
**Describe the bug**

When using BulkActions, if you click `SELECT ALL` then unselect one of the rows, it will seem like it's unselected but it is still selected.

**To Reproduce**

Anywhere the table is used, following the steps above will reproduce this behavior.

**Expected behavior**

When `SELECT ALL` is pressed, all rows' checkboxes could be disabled so they can't be unselected. That way, the user would have to click the `X` in the BulkAction bar to cancel the selection.

**Screenshots**

![image](https://user-images.githubusercontent.com/22064061/59450099-d9edb080-8dde-11e9-9968-7e361a65b606.png)

cmdalbem:
makes sense @rafarubim

kevinch:
Closing. This has already been fixed.

PR #690, opened 2019-06-13
Fix/select all reversible

rafarubim:

Disable all bulk action checkboxes when `SELECT ALL` is pressed

The `SELECT ALL` button should select all items in the table, even those who don't show in the current page or haven't been fetched. After the button is pressed, no row should be unselected since that won't really won't affect the flag `allLinesSelected`, which is `true`. This solves the following issue:
https://github.com/vtex/styleguide/issues/688

Go to
https://rafarefactor--bonati.myvtex.com/totalexpress/home?__disableSSR
and try selecting any item in the table, then click `SELECT ALL`. The checkboxes should be disabled.

image

#### Types of changes

### DIFF
--- a/react/components/Table/CheckboxContainer.js
+++ b/react/components/Table/CheckboxContainer.js
@@ -6,6 +6,7 @@ import Checkbox from '../Checkbox'
 class CheckboxContainer extends Component {
   static defaultProps = {
     partial: false,
+    disabled: false,
   }
 
   static propTypes = {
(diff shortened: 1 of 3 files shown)


react/components/Table/CheckboxContainer.js
```diff
@@ -12,11 +12,12 @@ class CheckboxContainer extends Component {
     checked: PropTypes.bool.isRequired,
     partial: PropTypes.bool,
     onClick: PropTypes.func.isRequired,
+    disabled: PropTypes.bool,
```
kevinch: Add this prop in defaultProps.

react/components/Table/index.js
```diff
@@ -213,6 +213,7 @@ class Table extends PureComponent {
               )}
               onClick={() => this.handleSelectLine(rowData)}
               id={rowData.id}
+              disabled={this.state.allLinesSelected}
```
kevinch: Use the `allLinesSelected` defined in `const {...} = this.state`

PR #691, opened 2019-06-13
Feature/bulkactions onchange

rafarubim:

Add bulkActions.onChange prop to Table component

With that prop, the user can pass a callback to the table component which is called every time there's a change in the selected rows, receiving the same parameters as `main's` and `others'` `handleCallback`.

Go to
https://rafarefactor--bonati.myvtex.com/totalexpress/home?__disableSSR
and try selecting some rows. You'll see that below the table there's a json string updating in real time containing all selected rows

image

#### Types of changes

### DIFF
--- a/react/components/Table/index.js
+++ b/react/components/Table/index.js
@@ -97,9 +97,18 @@ class Table extends PureComponent {
     )
   }
 
+  handleSelectionChange = () => {
+    if (this.props.bulkActions && this.props.bulkActions.onChange) {
+      const selectedParameters = this.state.allLinesSelected
+        ? { allLinesSelected: true }
+        : { selectedRows: this.state.selectedRows }
+        this.props.bulkActions.onChange(selectedParameters)
+    }
+  }
+
   handleSelectAllLines = () => {
     const { items: selectedRows } = this.props
-    this.setState({ selectedRows, allLinesSelected: true })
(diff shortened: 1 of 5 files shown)


react/components/Table/index.js
```diff
@@ -97,9 +97,18 @@ class Table extends PureComponent {
...
+        ? { allLinesSelected: true }
+        : { selectedRows: this.state.selectedRows }
+        this.props.bulkActions.onChange(selectedParameters)
+    }
+  }
+
   handleSelectAllLines = () => {
     const { items: selectedRows } = this.props
-    this.setState({ selectedRows, allLinesSelected: true })
+    this.setState({ selectedRows, allLinesSelected: true }, this.handleSelectionChange)
```
kevinch: Is it the same as `this.setState({ selectedRows, allLinesSelected: true }, () => this.handleSelectionChange)`?

react/components/Table/index.js
```diff
@@ -413,6 +422,7 @@ Table.propTypes = {
       allRowsSelected: PropTypes.func.isRequired,
     }),
     totalItems: PropTypes.number,
+    onChange: PropTypes.func,
```
kevinch: An example in the readme would be nice.

react/components/Table/index.js
rafarubim: Yes, but without using an anonymous function

PR #722, opened 2019-07-16
Add onExited prop

mariana-mendes:
as the title says. 

The prop `onExited` already exists in `ResponsiveModal` (react component internally used). Just passed the prop to parent component ( `ModalDialog`) as  `onCloseTransitionFinish`.
What `onCloseTransitionFinish` does? The function you pass to this prop will be fired when the Modal has exited and the animation is finished.

1. [Workspace](https://mariana--storecomponents.myvtex.com/admin/collections).

a

#### Types of changes

### DIFF
--- a/react/components/Modal/index.js
+++ b/react/components/Modal/index.js
@@ -17,6 +17,7 @@ class Modal extends PureComponent {
   }
 
   componentDidMount() {
+    console.log('altered')
     this.setShadowState(this.contentContainerReference.current)
   }
 
@@ -56,6 +57,7 @@ class Modal extends PureComponent {
       responsiveFullScreen,
       showTopBar,
       showBottomBarBorder,
+      onExited,
     } = this.props
     const { shadowBottom, shadowTop } = this.state
 
@@ -64,6 +66,7 @@ class Modal extends PureComponent {
         open={isOpen}
         little={centered}
         onClose={onClose}
+        onExited={onExited}
         closeOnEsc={closeOnEsc}
         closeOnOverlayClick={closeOnOverlayClick}
         showCloseIcon={showTopBar ? false : showCloseIcon}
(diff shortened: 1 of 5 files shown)


react/components/Modal/index.js
```diff
@@ -64,6 +65,7 @@ class Modal extends PureComponent {
         open={isOpen}
         little={centered}
         onClose={onClose}
+        onExited={onExited}
```
luciannojunior:
```suggestion
        onExit={onExit}
```

Change to `onExit` to maintain consistence

mariana-mendes:
> please rename it to `onExit`
> 
> (and by the way, what is the difference between `exit` and `close`?)

onExited is fired when the animation is finished.

react/components/Modal/index.js
mariana-mendes: https://react-responsive-modal.leopradel.com/#props

lbebber:
@mariana-mendes ahh I got it now. On the context of `react-transition-group` the name makes more sense, but on isolation here it's not very clear I think. 

How about `onCloseTransitionFinish` or something of the sort?

mariana-mendes:
> @mariana-mendes ahh I got it now. On the context of `react-transition-group` the name makes more sense, but on isolation here it's not very clear I think.
> 
> How about `onCloseTransitionFinish` or something of the sort?

@lbebber I think is good. But should I rename `handleExit = () => ... ` too?. 
`handleCloseTransitionFinish` seems  weird, what you think?

lbebber:
@mariana-mendes I don't see any problem with `handleCloseTransitionFinish`, except that it's maybe a bit too verbose, but I don't really mind if it's kept as `handleExit` because it's internal anyway.

react/components/Modal/index.js
```diff
@@ -162,6 +164,7 @@ Modal.propTypes = {
   responsiveFullScreen: PropTypes.bool,
   /** If true, show top bar with title */
   showTopBar: PropTypes.bool,
+  onCloseTransitionFinish: PropTypes.func,
```
lbebber:
```suggestion
  /** Event fired when the closing transition is finished */
  onCloseTransitionFinish: PropTypes.func,
```

Issue #552, opened 2019-02-19
RadioGroup text overpassing its box


italotabatinga:
**Describe the bug**

The RadioGroup text is overpassing the surrounding box when the text is a big word (isn't a phrase that can be divided in little words).

**To Reproduce**

1. Go to [VTEX Styleguide](https://styleguide.vtex.com/#/Components/Forms/RadioGroup)
2. Open Inspect Drawer
3. Expand the Drawer size so that the webpage width is minimum.

<img width="145" alt="screen shot 2019-02-19 at 4 42 42 pm" src="https://user-images.githubusercontent.com/26533708/53042747-be445000-3465-11e9-8d3f-8eae38b8eab7.png">

**Expected behavior**
I think the expected behavior would be to break the word so that it fits on the RadioGroup box.
<img width="139" alt="screen shot 2019-02-19 at 4 47 33 pm" src="https://user-images.githubusercontent.com/26533708/53042934-18451580-3466-11e9-9dec-54c5c8e19c2b.png">

**Screenshots**

The example above was an easy to reproduce one but not a real case, below is my real example, RadioGroup texts don't fit on iPhone 5/SE/6/7/8 viewports.
<img width="306" alt="screen shot 2019-02-19 at 4 53 10 pm" src="https://user-images.githubusercontent.com/26533708/53043305-f0a27d00-3466-11e9-89f3-86273f2a6baa.png">

After adding this style option `word-break: break-all` on the label components it seemed to work for me.
<img width="307" alt="screen shot 2019-02-19 at 4 53 36 pm" src="https://user-images.githubusercontent.com/26533708/53043309-f1d3aa00-3466-11e9-8bde-d7b89c78971c.png">

PR #726, opened 2019-07-18
Add break-word on label

kevinch:
Add break-word css prop to Radio label.

Fixes https://github.com/vtex/styleguide/issues/552

Checkout and resize browser window.

Before:
Screen Shot 2019-07-18 at 11 58 08

After:
Screen Shot 2019-07-18 at 11 58 00

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [8.66.1] - 2019-07-22
+
 ### Added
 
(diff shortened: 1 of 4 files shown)


cmdalbem:
@kevinch I'm not sure the solution you proposed is the best. It does fix the problem for this URL case, but for all other regular cases using "normal words" it breaks them in half. Notice in the example below how the words `should`, `thumbs` and `least` are broken in a weird way:

image

kevinch:
@cmdalbem agreed, i did a test too with both text and url . and the `break-word` works best. I was afraid it would not be able to break urls properly.

Screen Shot 2019-07-22 at 10 49 36

PR #727, opened 2019-07-19
Add disabled state for toolbar options

kevinch:
Add disabled state for toolbar upload & download options.

Download button has to be disabled in my app when there is nothing to be downloaded.

Checkout and add `disabled: true` locally.

Enabled:
Screen Shot 2019-07-19 at 15 10 11

Disabled:
Screen Shot 2019-07-19 at 15 10 02

Both:
Screen Shot 2019-07-19 at 15 12 41

#### Types of changes

### DIFF
--- a/react/components/Table/Toolbar.js
+++ b/react/components/Table/Toolbar.js
@@ -129,6 +129,8 @@ class Toolbar extends PureComponent {
       size: 'small',
     }
 
+    const forcedColor = 'c-on-base'
+
     return (
       <div
         id="toolbar"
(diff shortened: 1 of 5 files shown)


react/components/Table/Toolbar.js
```diff
@@ -270,17 +272,20 @@ class Toolbar extends PureComponent {
...
                 }
-                disabled={loading}
                 variation="tertiary"
+                disabled={download.disabled}
                 isLoading={download.isLoading}
                 size="small"
                 onClick={download.handleCallback}>
                 {download.label && (
-                  <span className="c-on-base">{download.label}</span>
+                  <span className={`${download.disabled ? '' : forcedColor}`}>
```
ericreis: Why do you need to force the color here? Wouldn’t the disable in the `ButtonWithIcon` component set this color?

kevinch: It's forced as a design decision. @cmdalbem might clarify this point.

ericreis: LGTM then :sparkles:

PR #736, opened 2019-07-30
Feature/modal container

rafarubim:

Add a workaround for using modal inside iframes

When the modal is inside one iframe, it only shades the screen inside the iframe, not all of the browser window, as the design is meant to do. This is solved by adding a "container" prop to the modal, which lets the user set where it will be rendered. Therefore, it can be used with `container={window.top.document.body}` to make it render outside the iframe.

You can test a version without the workaround here:
https://rafaprtest4--bonati.myvtex.com/admin/example

And the workaround is being used here:
https://rafaprtest3--bonati.myvtex.com/admin/example

Modal before:
image

Modal after applying the workaround:

image

#### Types of changes

### DIFF
--- a/react/components/Modal/index.js
+++ b/react/components/Modal/index.js
@@ -63,7 +63,7 @@ class Modal extends PureComponent {
     return (
       <ResponsiveModal
         open={isOpen}
-        little={centered}
+        center={centered}
         onClose={onClose}
         onExited={onCloseTransitionFinish}
         closeOnEsc={closeOnEsc}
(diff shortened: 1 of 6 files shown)


react/components/Modal/index.js
```diff
@@ -143,6 +145,8 @@ Modal.propTypes = {
   children: PropTypes.node.isRequired,
   /** Center the modal (for small content) */
   centered: PropTypes.bool,
+  /** Container in which the modal is rendered */
+  container: PropTypes.object,
```
ericreis: What happens if you don't this prop?

rafarubim: It defaults to a <div> appended to `document.body` by the library behind
https://github.com/pradel/react-responsive-modal/blob/master/src/modal.js#L69

PR #743, opened 2019-08-08
Adding prop and update the state when component updates

laerc:

Here we will add a prop called `selectedRows` that contains the information about the selected items. If this prop is not `undefined`, then the selectedRows will be controlled outside the component(by the caller), if `undefined` the table will control the selected elements.

When you need to control the deselect some items in the table when an event happens, this is not possible with the current props.

1. Go to https://laercio--basedevmkp.myvtex.com/admin/suggestion/pending
2. Select some items using the bulk actions(1 one or more)
3. Then Choose `Approve as new product`
4. Close the modal(yes, just close it)
5. Then the items should be unselected.

image
image
image

#### Types of changes

### DIFF
--- a/react/components/Table/index.js
+++ b/react/components/Table/index.js
@@ -31,9 +31,14 @@ class Table extends PureComponent {
     }
   }
 
-  componentDidUpdate(prevProps) {
-    if(this.props.bulkActions && this.props.bulkActions.selectedRows) {
-      this.setState({selectedRows: this.props.bulkActions.selectedRows})
+  componentDidUpdate() {
+    const { bulkActions } = this.props
+    const { selectedRows } = this.state
+
+    if(bulkActions 
+      && bulkActions.selectedRows
+      && bulkActions.selectedRows != selectedRows) {
+      this.setState({selectedRows: bulkActions.selectedRows})
     }
   }
 
--- a/react/components/Table/index.js
+++ b/react/components/Table/index.js
@@ -35,9 +35,9 @@ class Table extends PureComponent {
     const { bulkActions } = this.props
     const { selectedRows } = this.state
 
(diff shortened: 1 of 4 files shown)


react/components/Table/index.js
```diff
@@ -26,11 +26,17 @@ class Table extends PureComponent {
...
       allChecked: false,
-      selectedRows: [],
+      selectedRows: this.props.bulkActions.selectedRows || [],
       allLinesSelected: false,
     }
   }
 
+  componentDidUpdate(prevProps) {
+    if(this.props.bulkActions && this.props.bulkActions.selectedRows) {
+      this.setState({selectedRows: this.props.bulkActions.selectedRows})
```
rafarubim: Won't this call to `setState` in `componentDidUpdate` cause an endless loop? I'm feeling a lag when the bulkActions are used. Maybe this is caused by the fact that react limits the amount of renders. It wouldn't be stuck in an infinite loop but it would be the cause of this lag, since there's a lot of rerender happening.

laerc: Me and @rafarubim have discussed and tested this part and we found that is not a problem. Furthermore, we add a new check(prop and state arrays are different) to prevent the component of re-rendering.

react/components/Table/index.js
```diff
@@ -26,11 +26,22 @@ class Table extends PureComponent {
...
     }
   }
 
+  componentDidUpdate() {
+    const { bulkActions } = this.props
+    const { selectedRows } = this.state
+
+    if(bulkActions 
+      && bulkActions.selectedRows
+      && bulkActions.selectedRows != selectedRows) {
```
kevinch: Use absolute comparison `!==`

PR #750, opened 2019-08-14
Added Toolbar, ButtonGroup and DensityButton to Table v2

matheusps:

🎣Added density prop to `useTableState`
👻`Toolbar` to hold buttons and inputs.
👻`ButtonGroup` to organize all buttons of the toolbar (including their order).
👻Density toggle button to allow switching between different densities.

Add different densities for the table rows.

Run locally with `yarn && yarn start`

#### Types of changes


react/components/EXPERIMENTAL_Table/Toolbar/Menu/Item.tsx
```diff
@@ -0,0 +1,37 @@
...
+  isSelected,
+  handleCallback,
+  closeMenuOnClick,
+  children,
+}) => {
+  const { setBoxVisible } = useMenuContext()
+
+  const handleClick = () => {
+    closeMenuOnClick && setBoxVisible(false)
+    handleCallback()
```
kevinch: Will it be always available?

react/components/EXPERIMENTAL_Table/index.tsx
```diff
@@ -18,20 +19,29 @@ const propTypes = {
   items: PropTypes.arrayOf(PropTypes.object),
   isEmpty: PropTypes.bool,
   tableHeight: PropTypes.number,
+  rowHeight: PropTypes.number,
+  selectedDensity: PropTypes.oneOf(['low', 'medium', 'high']),
```
kevinch: Could those value be called from an external const? I've seen those written already somewhere else.

react/components/EXPERIMENTAL_Table/Toolbar/Menu/Item.tsx
matheusps: yes, always

react/components/EXPERIMENTAL_Table/index.tsx
matheusps: The states are controlled by the `useTableState` hook that is used outside of the `Table` itself. So, the parent has full control of these values before passing them as props. It is very useful if the user wants to take actions based on these values. Do you agree with the usage or that is best to call the hook inside of `Table` component  (this will reduce the props to `columns`, `items` and `density`)?

matheusps: I find very useful the values outside, because instead of overloading the table with callback props such as `onToggleDensity`, `onHiddeFields`, `setSetDensityToLow`, `onWhatever` I can just use an `effect` for it. Such as: 

```js
const tableState = useTableState(...);

useEffect( ( ) => {
  if (tableState.selectedDensity === 'low') {  /* do my magic */ }
}, [tableState.selectedDensity])

<Table {...tableState} />
```
It's verry powerful and avoids 50 milion abstractions.
And I do think that a prop called `state` that receives all the states at the table, would look better (this spreading is not that appealing and can cause confusion). Such as:  `<Table state={tableState} />`

Would like to hear your thoughts on this. 🙂

react/EXPERIMENTAL_Table.ts
```diff
@@ -1,3 +1,5 @@
 import Table from './components/EXPERIMENTAL_Table/index'
+import { DENSITIES } from './components/EXPERIMENTAL_Table/constants'
```
kevinch: Is it required to import/export this here?

react/components/EXPERIMENTAL_Table/index.tsx
kevinch: Can you do this? `selectedDensity: PropTypes.oneOf(DENSITY_OPTIONS),`

matheusps: Done

react/EXPERIMENTAL_Table.ts
matheusps: This allows the developer to access these values so that different densities are treated consistently throughout the code.

react/components/EXPERIMENTAL_Table/hooks/useDensity.ts
```diff
@@ -0,0 +1,17 @@
+import { useState, useMemo } from 'react'
+import { getRowHeight } from '../util'
+
+const useDensity = (density: Density) => {
+  const [selectedDensity, setSelectedDensity] = useState<Density>(density)
+  const rowHeight = useMemo(() => getRowHeight(selectedDensity), [
```
lucasecdb: I don't think this memo is necessary, since the `getRowHeight` function isn't CPU-intensive.

be careful to not overuse the `useMemo` hook, since we are "paying" with a higher memory usage, for simple functions or functions that doesn't affect the rendering of the components we can call the function directly.

matheusps: solved

PR #774, opened 2019-08-22
Replace min-h-100 with vh version

kaisermann:

Fix the layout in admin pages with content smaller than the window height.

Currently, the `<Layout />` component uses `min-height: 100%` instead of `min-height: 100vh`. The `Layout` will only have the window height (100%) if its parent has a defined height, which it doesn't.

Just go to an admin page and change the class of a layout component to `min-vh-100`.

**Before**:
image

**After**: 
image

#### Types of changes


ohmyguigs:
Oh! and don't forget the changelog!

PR #805, opened 2019-08-29
Move appearance inline style to class

lucasecdb:
Move the `appearance: none` from an inline style to a CSS class in the `Input` component.

In AMP pages, the prefix on the `appearance` property is prohibited, and renaming it to just `appearance` would probably make it not work in the browsers that only support the prefixed version, so moving it to a class and letting postcss add the prefix fixes the issue.

#### Types of changes

### DIFF
--- a/react/components/Input/Input.css
+++ b/react/components/Input/Input.css
@@ -1,4 +1,4 @@
-/* Edge-specific */
+/* removes edge clear button on input */
 .hideDecorators::-ms-clear{
   display: none;
 }
--- a/react/components/Input/Input.css
+++ b/react/components/Input/Input.css
@@ -1,3 +1,4 @@
+/* Edge-specific */
 .hideDecorators::-ms-clear{
   display: none;
 }
--- a/react/components/Input/edge.css
+++ b/react/components/Input/edge.css
@@ -1,3 +1,7 @@
 .hideDecorators::-ms-clear{
   display: none;
 }
+
+.noAppearance {
+  appearance: none;
+}
--- a/react/components/Input/index.js
+++ b/react/components/Input/index.js
@@ -1,7 +1,7 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 
-import { hideDecorators } from './edge.css'
+import styles from './edge.css'
 import { withForwardedRef, refShape } from '../../modules/withForwardedRef'
 
 class Input extends Component {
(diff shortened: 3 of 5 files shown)


klzns:
Can you provide a linked workspace? So we can check what postcss did.

lucasecdb:
@klzns it's linked [here](https://lucas--delivery.myvtex.com).

react/components/Input/index.js
```diff
@@ -97,7 +97,7 @@ class Input extends Component {
     let prefixClasses = `${borderRadiusBase} br-0 br--left `
     let suffixClasses = `${borderRadiusBase} bl-0 br--right `
 
-    let classes = `${widthClass} ${box} ${hideDecorators} ${borderRadius} bn outline-0 `
+    let classes = `${widthClass} ${box} ${styles.hideDecorators} ${styles.noAppearance} ${borderRadius} bn outline-0 `
```
kevinch: Not sure this will scale fine if 20 or more styles are added in the css file.

lucasecdb:
> This will affect the component's look? Or it is just a technical improvement?

no, it's just a technical improvement.

lucasecdb:
@TasGuerciMaia @kevinch @klzns can you review?

react/components/Input/edge.css
```diff
@@ -1,3 +1,7 @@
 .hideDecorators::-ms-clear{
   display: none;
-}
\ No newline at end of file
+}
+
+.noAppearance {
```
kevinch: Is this designed only for IE Edge?

lucasecdb: I don't know what was the intention of the dev that added this property. I just want to remove the error on AMP pages as I've said in the PR description

kevinch: Agreed, but even though, if it does not belong to an Edge fix it should not appear within this file dedicated to css for Edge.

lucasecdb: you want me to move this to a `Input.css` file?

kevinch: Sounds good to keep the project organised. I also made another comment above.

react/components/Input/index.js
lucasecdb: any ideas?

kevinch: Could you do a loop of some kind?

lucasecdb: a loop? to add all classes in `styles`? I don't think it'll be very readable

react/components/Input/edge.css
lucasecdb: why we have a stylesheet specific to edge bugs? I think it makes more sense to rename this `edge.css` to `Input.css` and add a comment to the `hideDecorators` class. When I first saw this class I had no idea it was related to the Edge browser.

kevinch: This could be a solution indeed.

react/components/Input/index.js
lucasecdb: since this is not really related with the changes I made in this PR, I think it's best to leave this issue for another PR

react/components/Input/Input.css
```diff
@@ -0,0 +1,8 @@
+/* Edge-specific */
```
kevinch: Update/remove comment.

lucasecdb: done :-)

PR #812, opened 2019-09-02
Add ButtonColumns to EXPERIMENTAL_TABLE_V2

matheusps:
➕ `ButtonColumns` to `EXPERIMENTAL_TableV2`

The `Table v2` toolbar may toggle columns visibility.

`yarn start`?

Screen Shot 2019-09-02 at 15 58 53 (2)

#### Types of changes

### DIFF
--- a/react/components/EXPERIMENTAL_Table/Toolbar/ButtonColumns.tsx
+++ b/react/components/EXPERIMENTAL_Table/Toolbar/ButtonColumns.tsx
@@ -1,0 +1,74 @@
+import React, { useMemo, FC } from 'react'
+
+import Toggle from '../../../Toggle'
+import IconColumns from '../../icon/Columns/index'
+import Menu from './Menu/index'
+
+import { ICON_SIZE, COLUMNS_BOX, NAMESPACES } from '../constants'
+import useTableContext from '../hooks/useTableContext'
+
+export type ButtonColumnsProps = {
+  label: string
+  showAllLabel: string
+  hideAllLabel: string
+  alignMenu: Alignment
+  disabled: boolean
+}
+
+const ButtonColumns: FC<ButtonColumnsProps> = ({
+  label,
+  showAllLabel,
+  hideAllLabel,
+  alignMenu,
+  disabled,
+}) => {
+  const {
+    hiddenColumns,
+    columns,
+    hideAllColumns,
+    showAllColumns,
+    toggleColumn,
+  } = useTableContext()
+
+  const calculateFieldsBoxHeight = () => {
+    const estimate = Object.keys(columns).length * COLUMNS_BOX.ITEM_HEIGHT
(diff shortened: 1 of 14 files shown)


react/components/EXPERIMENTAL_Table/hooks/useHiddenColumns.ts
```diff
@@ -0,0 +1,41 @@
...
+  const [hiddenColumns, setHiddenColumns] = useState(getHiddenColumns(columns))
+
+  const visibleColumns = useMemo(() => {
+    const reducer = (acc: Array<string>, col: Column) =>
+      hiddenColumns.includes(col.id) ? acc : [...acc, col]
+
+    return columns.reduce(reducer, [])
+  }, [hiddenColumns, columns])
+
+  const toggleColumn = (id: string) => {
```
lucasecdb: I think it's good practice to *always* memoize callbacks and values from custom hooks, so the component that uses them can safely add them to the dependencies of an `useEffect`

react/EXPERIMENTAL_Table.ts
```diff
@@ -1,7 +1,5 @@
 import Table from './components/EXPERIMENTAL_Table/index'
-import useTableState from './components/EXPERIMENTAL_Table/hooks/useTableState'
 import { TABLE_DENSITIES } from './components/EXPERIMENTAL_Table/constants'
 
 export default Table
-export { useTableState }
 export { TABLE_DENSITIES }
```
vitoriaHeliane: The `TABLE_DENSITIES` should be exported in the same way that the `useTableState` is.

matheusps: Solved on another PR

react/components/EXPERIMENTAL_Table/hooks/useHiddenColumns.ts
matheusps: ✅

react/components/EXPERIMENTAL_Table/Toolbar/ButtonColumns.tsx
```diff
@@ -0,0 +1,74 @@
...
+    hiddenColumns,
+    columns,
+    hideAllColumns,
+    showAllColumns,
+    toggleColumn,
+  } = useTableContext()
+
+  const calculateFieldsBoxHeight = () => {
+    const estimate = Object.keys(columns).length * COLUMNS_BOX.ITEM_HEIGHT
+    return estimate > COLUMNS_BOX.MAX_HEIGHT ? COLUMNS_BOX.MAX_HEIGHT : estimate
```
gugabribeiro: Instead of do this ternary comparison it is enough to return the minimum between both values. Something like `Math.min(estimate, COLUMNS_BOX.MAX_HEIGHT)` or even `Math.min(Object.keys(columns).length * COLUMNS_BOX.ITEM_HEIGHT, COLUMNS_BOX.MAX_HEIGHT)`

Issue #823, opened 2019-09-05
Kill gh-pages deploy in branch 8.x


ohmyguigs:

kevinch:
For some reason the keyword here did not work.

PR #824, opened 2019-09-05
Remove deploy task

kevinch:
Closes https://github.com/vtex/styleguide/issues/823

A merge in 8.x would trigger a deploy of the Styleguide site with the 8x branch.

#### Types of changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "clean:lib": "rm -rf lib",
     "clean:modules": "rm -rf modules",
     "compile": "run-s clean:lib clean:modules build:types babel:lib babel:modules",
-    "prepublishOnly": "run-s compile deploy",
+    "prepublishOnly": "run-s compile",
     "postreleasy": "vtex publish --public --verbose",
     "lint": "eslint react",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -15,8 +14,6 @@
     "babel:modules": "NODE_ENV=production babel ./react/modules --out-dir ./modules --ignore '__tests__,*.spec.js' --ignore '*.md' --copy-files",
     "clean:lib": "rm -rf lib",
     "clean:modules": "rm -rf modules",
-    "compile": "run-s clean:lib clean:modules build:types babel:lib babel:modules",
```
ohmyguigs: doesn't this one need to stay? I guess npm publish uses this one

package.json
```diff
@@ -15,8 +14,6 @@
...
-    "prepublishOnly": "run-s compile deploy",
```
ohmyguigs: should be just
`"prepublishOnly": "run-s compile",`

package.json
kevinch: Only used for `prepublishOnly`. Could it be used somewhere else?

kevinch: Reverted.

PR #832, opened 2019-09-09
Do not show column for showing/hiding when there is to text

ericreis:
Do not show columns in the show/hide column list if there is no text (trimmed) for that column (title or field name).

The following weird behavior:
image

image


cmdalbem:
Design-wise it's 👍

PR #833, opened 2019-09-10
Add the autocomplete input component

estacioneto:

Add the `AutocompleteInput` component.

There was only a Figma autocomplete but no autocomplete implemented in our Styleguide's react components.

Clone this repo, checkout to this branch then run `yarn && yarn styleguide`

##### Component API

image

##### Component visual

AutocompleteInput

#### Types of changes


react/components/AutocompleteInput/icon/ClearInput.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
+
+interface IconProps {
+  size?: number
+  color?: string
+}
+
+const ClearInputIcon: React.FC<IconProps> = ({
```
vitoriaHeliane: Use the `ClearIcon` from the `styleguide`?

ericreis:
I think this component is really similar to the **Select** component. How is this component different from the Select? We need to understand those differences before starting the development process.

estacioneto:
@ericreis, they really are quite similar, but there are several semantic differences:

- Select is meant to be a component where you can filter the options but the user **must** select a given option. The autocomplete just suggests options that can be ignored.
- In general, autocompletes are used in the search context. Selects in a form context. I think this is the major difference, btw.
- ~Using the Select component to implement the autocomplete doesn't smell good to me, tbh… We have a Select component and a Dropdown component and they are quite similar too but for different purposes, so…~
- All selectable options should be accessible even if the user doesn't type anything in the input, but the autocomplete suggestions should be shown only based on the user's interaction (or showing last searched terms, for example).

Anyway, I'd like to see the other's opinion about this too. All I've said here was based on a search on some styleguides/style-frameworks on the web (e.g.: [Angular Material](https://material.angular.io), [Semantic UI](https://semantic-ui.com))

TasGuerciMaia:
Well they are quite similar. Although I agree with the differences @estacioneto shared. This component is really important to be used along with the table component, for example. I believe that the select component returns options from a pre select list. In the other hand the auto complete input helps the user on broad searches.

Yet I have a question. 
- How it retrieves the information to complete? Is it from the backend?

estacioneto:
Yes, @TasGuerciMaia . In general, the information from the autocomplete comes from the server, for example, in the [admin-collections](https://github.com/vtex/admin-collections/blob/master/react/components/CollectionsTable/CollectionsAutocomplete/index.tsx#L50) it's used to filter the collections' names and product's names. We fetch the products with names that match the input value and then we show them in the autocomplete popup.

react/components/AutocompleteInput/README.md
```diff
@@ -0,0 +1,255 @@
+#### AutocompleteInput is a search input which, enhanced with a suggestions pop-up, gives the user a more complete search experience when needed.
+
+### 👍 Dos
+
+- Help the user avoid errors.
+- Semantically similar phrases or words but grammarly different should appear once. E.g.: "Dress" and "Dresses".
+- Highlight the suggestions.
+
```
matheusps: Maybe this, instead of the `Dont's`. What do you think? 🤔

```suggestion
- Keep the number of suggestions between 6 and 10.
```

react/components/AutocompleteInput/README.md
```diff
@@ -0,0 +1,255 @@
...
+    placeholder: 'Search user... (e.g.: Ana)',
+    value: term,
+  }
+  return <AutocompleteInput input={input} options={options} />
+}
+
+;<UsersAutocomplete />
+```
+
+#### Show search history
```
matheusps:
```suggestion
#### Show search history.
```

react/components/AutocompleteInput/Option.tsx
```diff
@@ -0,0 +1,91 @@
...
+}
+
+const Option: React.FunctionComponent<
+  PropTypes.InferProps<typeof propTypes>
+> = props => {
+  const { icon, selected, roundedBottom, searchTerm, key, onClick } = props
+  const [highlightOption, setHighlightOption] = useState(false)
+  const value = getTermFromOption(props.value)
+
+  const renderOptionValue = (): string | React.ReactElement => {
```
vitoriaHeliane: It could be a `Component`.

react/components/AutocompleteInput/Option.tsx
```diff
@@ -0,0 +1,91 @@
...
+      onClick={onClick}>
+      <span className="h1 flex items-center">
+        <span className="mr3 c-muted-2 flex pt1">{icon}</span>
+        {renderOptionValue()}
+      </span>
+    </button>
+  )
+}
+
+Option.propTypes = propTypes
```
vitoriaHeliane: If you are using generics, why is it necessary?

react/components/AutocompleteInput/SearchInput/style.css
```diff
@@ -0,0 +1,7 @@
+.br--bottom-left-0 {
```
vitoriaHeliane: You could use `br--top` class from the `Tachyons`, so it will not be necessary.

react/components/AutocompleteInput/SearchInput/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+  const handleBlur = (e: React.FocusEvent<HTMLInputElement>) => {
+    setFocused(false)
+    onBlur && onBlur(e)
+  }
+
+  const activeClass = classnames({
+    'b--muted-3': focused,
+    'b--muted-4': !focused,
+    [`${style['br--bottom-left-0']} ${style['br--bottom-right-0']}`]: !roundedBottom,
```
vitoriaHeliane: You could use `br--top` class from the `Tachyons`.

react/components/AutocompleteInput/SearchInput/style.css
```diff
@@ -0,0 +1,7 @@
...
+  border-bottom-left-radius: 0;
+}
+
+.br--bottom-right-0 {
```
vitoriaHeliane: Same here.

react/components/AutocompleteInput/Option.tsx
```diff
@@ -0,0 +1,91 @@
...
+  PropTypes.string,
+  structuredAutocompleteOptionShape,
+])
+ 
+export const getTermFromOption = (option: AutocompleteOption): string =>
+  typeof option === 'string' ? option : option.label
+
+const propTypes = {
+  /** Determine if the option should have a rounded bottom */
+  roundedBottom: PropTypes.bool,
```
kevinch: Is that a UI feature?

react/components/AutocompleteInput/Option.tsx
```diff
@@ -0,0 +1,91 @@
...
+    return (
+      <span className="truncate">
+        <span className="fw7">{prefix}</span>
+        {match}
+        <span className="fw7">{suffix}</span>
+      </span>
+    )
+  }
+
+  const buttonClasses = classnames('bn  w-100 tl pointer pa4 f6', {
```
kevinch: Nitpick:
```suggestion
  const buttonClasses = classnames('bn w-100 tl pointer pa4 f6', {
```

react/components/AutocompleteInput/Option.tsx
```diff
@@ -0,0 +1,91 @@
...
+    'br2 br--bottom': roundedBottom,
+    'bg-muted-5': highlightOption || selected,
+    'bg-base': !highlightOption && !selected,
+  })
+
+  return (
+    <button
```
kevinch: Cant it be the `<Button>` from Styleguide?

react/components/AutocompleteInput/README.md
```diff
@@ -0,0 +1,255 @@
...
+
+### 👍 Dos
+
+- Help the user avoid errors.
+- Semantically similar phrases or words but grammarly different should appear once. E.g.: "Dress" and "Dresses".
+- Highlight the suggestions.
+
+### 👎 Don'ts
+
+- 6 is a great number of suggestions. Don't exceed 10 suggestions.
```
kevinch:
```suggestion
- Exceed 10 suggestions. 6 is a great number of suggestions.
```

react/components/AutocompleteInput/SearchInput/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+import IconSearch from '../../icon/Search'
+
+import ClearInputIcon from '../../icon/Clear'
+
+const propTypes = {
+  /** Determine if the input's bottom corners should be rounded or not */
+  roundedBottom: PropTypes.bool,
+  /* Input props */
+
```
kevinch: Missing prop?

react/components/AutocompleteInput/Option.tsx
estacioneto: Yes. It can be renamed to `lastOption`, if you think it's better...

react/components/AutocompleteInput/Option.tsx
estacioneto: No, it can't. The Styeguide button doesn't support this level of styling

react/components/AutocompleteInput/SearchInput/index.tsx
estacioneto: Actually, it's just to comment that these props are related to the input element itself. I'll make it closer to the `value` prop documentation.

estacioneto:
Então, @oldufre, deveria também utilizar o `caption` para o texto do input? Eu acho que fica muito complicado, não? Por isso deixei apenas o texto principal ser utilizado. Tanto que no histórico de pesquisa apareceu qual delas foi selecionada. O que vc acha?

PR #842, opened 2019-09-18
Set Tooltip z-index to avoid being shadowed by other components

estacioneto:

As the title says.

image

#### Types of changes

### DIFF
--- a/react/components/utils/zIndex.ts
+++ b/react/components/utils/zIndex.ts
@@ -1,0 +1,4 @@
+export default {
+  modal: 10000,
+  tooltip: 11000,
+}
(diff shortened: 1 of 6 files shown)


react/components/utils/zIndex.ts
```diff
@@ -0,0 +1,4 @@
+export default {
+  modal: 10000,
+  tooltip: 11000,
```
kevinch: Not sure though that those values have to be this high.

estacioneto: Me neither. Actually, we could follow something [like this](https://material-ui.com/customization/z-index/). I didn't change the modal `zIndex` because I don't really know if something would break.

kevinch: Interesting, it could be of some help yes.

PR #843, opened 2019-09-18
Removes unnecessary margins and paddings from hideBorder

cmdalbem:
Also, updates the examples in the documentation so as to encourage usage of the component with the `hideBorder` ON.

## Before

image

## After

image

### DIFF
--- a/react/components/RadioGroup/index.js
+++ b/react/components/RadioGroup/index.js
@@ -22,7 +22,7 @@ class RadioGroup extends React.Component {
             <label
               className={`db br3 ${classNames({
                 'b--muted-4': !hideBorder,
-                ba: !hideBorder,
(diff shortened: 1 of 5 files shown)


react/components/RadioGroup/index.js
```diff
@@ -20,9 +20,11 @@ class RadioGroup extends React.Component {
           const id = `${name}-${i}`
           return (
             <label
-              className={`db pv2 ph4 br3 ${classNames({
+              className={`db br3 ${classNames({
                 'b--muted-4': !hideBorder,
                 ba: !hideBorder,
```
matheusps:
```suggestion
                'b--muted-4 ba pv2 ph4': !hideBorder,
```
You can put the classes together since the condition is the same.

Issue #789, opened 2019-08-29
Placeholder text in input is too dark


davicosta99:
**Describe the bug**
Placeholder text in input is too dark and, as so, may appear to have been typed.

**To Reproduce**
Open the styleguide documentation and see for yourself.

**Expected behavior**
It should be light enough for it to give the impression that it's blank. This doesn't create an acessibility issue, since crucial information should never be in it anyway. The use of placeholder text is very debated, [actually](https://www.nngroup.com/articles/form-design-placeholders/), but this is needed if we are to support it.

**Screenshots**
![Group 3](https://user-images.githubusercontent.com/8797476/63942437-25117a00-ca44-11e9-8de8-fac72edb041d.png)

PR #847, opened 2019-09-19
Add css placeholder

kevinch:
Closes https://github.com/vtex/styleguide/issues/789

Before:
Screen Shot 2019-09-19 at 12 38 29

After:
Screen Shot 2019-09-19 at 12 36 36

#### Types of changes

### DIFF
--- a/react/components/Input/Input.css
+++ b/react/components/Input/Input.css
@@ -10,17 +10,17 @@
 /* placeholders are using c-muted-1's value */
 ::-webkit-input-placeholder {
   /* Chrome/Opera/Safari */
-  color: #727273;
+  color: #979899;
 }
 ::-moz-placeholder {
   /* Firefox 19+ */
-  color: #727273;
+  color: #979899;
 }
 :-ms-input-placeholder {
   /* IE 10+ */
-  color: #727273;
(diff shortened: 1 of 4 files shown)


react/components/Input/Input.css
```diff
@@ -1,8 +1,26 @@
...
+}
+::-moz-placeholder {
+  /* Firefox 19+ */
+  color: #727273;
+}
+:-ms-input-placeholder {
+  /* IE 10+ */
+  color: #727273;
+}
+:-moz-placeholder {
```
kaisermann: You could transform all these rules into a single one: `	::-webkit-input-placeholder, ::-moz-placeholder, :-ms-input-placeholder, :-moz-placeholder { ... }`

kevinch: I should.

kevinch: Does not work:

> This is because a browser will only apply a rule form a selector it can fully interpret. For a webkit type browser -webkit-input-placeholder is valid but -moz-placeholder is not, so it trashes the entire selector, and vise-versa for a geeko based browser. The solution is to separate browser specific selectors.

Issue #848, opened 2019-09-19
Add data-attributes prop for e2e tests


kevinch:
Add the prop `data-attribute` or `data-cy` so that apps using components can pass attributes meant for e2e testing easily and prevent the creation of almost-useless `div` containers.

emersonlaurentino:
I believe it could be used `data-testid` as a convention to define elements that can be found by testing.

We can use the same attribute for testing library and cypress. So you can use the [testing library's getByTestId function](https://testing-library.com/docs/dom-testing-library/api-queries#bytestid) and keep it a good practice for cypress.

>  Best Practice: Use data-* attributes to provide context to your selectors and insulate them from CSS or JS changes.

Ref.: [Best Practices by Cypress](https://docs.cypress.io/guides/references/best-practices.html)

kevinch:
Good point. `data-cy` seems too focused on Cypress while `data-testid` is more generic.

PR #856, opened 2019-09-26
Add data-testid attribute

kevinch:
Add simple data attribute `testid` to prevent developers from adding an extra html element around the element to be tested. Usually via e2e tests.

Closes https://github.com/vtex/styleguide/issues/848

Checkout & add the attribute `testId` to one of the following components:
- `Input`
- `Dropdown`
- `EmptyState`
- `Button`
- `PageBlock`

#### Types of changes

### DIFF
--- a/react/components/Dropdown/index.js
+++ b/react/components/Dropdown/index.js
@@ -83,6 +83,7 @@ class Dropdown extends Component {
     const {
       label,
       id,
+      testId,
       value,
       size,
       disabled,
@@ -174,7 +175,8 @@ class Dropdown extends Component {
       <div
         className={`vtex-dropdown ${
           isInline ? 'vtex-dropdown--inline dib' : ''
-        }`}>
+        }`}
+        data-testid={testId}>
         <label>
           {label && <span className={labelClasses}>{label}</span>}
           <div className={containerClasses} style={containerStyle}>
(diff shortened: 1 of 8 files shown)


react/components/Input/index.js
```diff
@@ -182,7 +183,7 @@ class Input extends Component {
     }
 
     return (
-      <label className="vtex-input w-100">
+      <label className="vtex-input w-100" data-testid={testId}>
```
salesfelipe: I think this should be inserted on the input itself.

react/components/Dropdown/index.js
```diff
@@ -174,7 +175,8 @@ class Dropdown extends Component {
       <div
         className={`vtex-dropdown ${
           isInline ? 'vtex-dropdown--inline dib' : ''
-        }`}>
+        }`}
+        data-testid={testId}>
```
salesfelipe: I think this should go on the select

react/components/Input/index.js
kevinch: Thought so too but i think it's best at the root level. Developer can then use something like ```
cy.get('[data-testid=string]').within(result => {})
```
To look for label, input etc.

salesfelipe: If you are looking for a label you don't need the id, the id is best used to get the value of the input or if the input is disabled...

kevinch: You can still do a `cy.get('[data-testid=string] input')` and get that. You could also do a `cy.get('[data-testid=string] label')` to get the text for some reason.

kevinch: Updated. Does not make much sense to target other elements.

react/components/Dropdown/index.js
kevinch: Updated.

PR #864, opened 2019-09-30
feat(Divider): New component to split elements

### LINKED ISSUE #862: Divider component
## Overview

Divider is a simple component represent by a line do segregate visually two groups of content. It can be used vertically or horizontally as well.

## Purpose

Nowadays we use borders to divide blocks or list items in layouts. Semantically, they are not borders, but dividers. It's very common to use dividers in Uis (in admins and stores), so this component would be very useful.

## Context and Examples

Screen Shot 2019-09-30 at 11 56 45

emersonlaurentino:
closes #862

closes #862

Access the styleguid docs and test on Divider section

Captura de Tela 2019-09-30 às 14 12 54

#### Types of changes


cmdalbem:
@emersonlaurentino do we have usage for the color and size versions? I don't remember seeing any designs in VTEX with that need.

TasGuerciMaia:
I agree with @cmdalbem. In addition to it I believe we should always set a default spacing for dividers. It would help us to keep consistency.

emersonlaurentino:
@cmdalbem @TasGuerciMaia I dont know, do you think it is best to leave the size fixed at `1px` and `bg-light-gray` as the default color?

and what do you think about the `gap`?

cmdalbem:
> @cmdalbem @TasGuerciMaia I dont know, do you think it is best to leave the size fixed at `1px` and `bg-light-gray` as the default color?

Yep.

> and what do you think about the `gap`?

I tend to think we don't need this too. The spacing could be dealt by the elements around using the regular padding/margins functional classes.

emersonlaurentino:
@cmdalbem With the gap it is not necessary to add margin to the different elements.

react/components/Divider/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+  color: PropTypes.string,
+}
+
+const Divider: React.FC<PropTypes.InferProps<typeof propTypes>> = props => {
+  const size = {
+    [props.orientation === 'horizontal' ? 'height' : 'width']: props.size,
+  }
+  const gap =
+    props.orientation === 'horizontal'
+      ? `mt${props.gap} mb${props.gap}`
```
jgfidelis: why not `mv${props.gap}` and `mh${props.gap}` for ml and mr below?

jgfidelis: and you may want to validate props.gap, otherwise user can pass a value like: `gap: '0 my-custom-tachyon-token'` and insert any class he wants there hehe

emersonlaurentino: i don't know all tachyons classes yet, but we will not use the gap anymore.

TasGuerciMaia:
Well either we have gap in the component or not. It sounds awkward to me having both in the same divider component.

Issue #873, opened 2019-10-03
[Table] Bulk row selection on filtered data


ericreis:
**Describe the bug**
When we select table rows of a filtered table and then remove the filter, the selected rows change. 

As of this moment the table is selecting its data based on the row number showing on the screen. When we use filters (pagination, also) that "row number" changes causing the bug.

**To Reproduce**
1. Go to a table that has filters and bulk actions turned on
2. Execute a filter that changes the position of any row in the table
3. Select that row
4. Clear the filter (or change it in a way that the selected row "moves")
5. See that the selected row changes.

**Expected behavior**
Maintain the originally selected row.

**Screenshots**
![Oct-03-2019 15-15-23](https://user-images.githubusercontent.com/5971264/66152866-c31ed400-e5f0-11e9-943b-53ed06b86685.gif)

**Additional context**
The solution for this could (and should?) be extended to pagination.

PR #874, opened 2019-10-03
Feature/table bulk actions deselect lines on input search clear

ericreis:
Deselect all lines on input search clear.

Prevents one possible wrong behavior mentioned in #873.

Last example of http://10.1.14.209:6060/#/Components/Display/Table

N/A

### DIFF
--- a/react/components/Table/index.js
+++ b/react/components/Table/index.js
@@ -141,8 +141,13 @@ class Table extends PureComponent {
   }
 
   handleInputSearchClearWithBulkAction = event => {
+    const {
+      toolbar: {
+        inputSearch: { onClear = () => {} },
+      },
+    } = this.props
     this.handleDeselectAllLines()
-    this.props.toolbar.inputSearch.onClear(event)
+    onClear(event)
   }
 
   render() {
--- a/react/components/Table/index.js
+++ b/react/components/Table/index.js
@@ -280,8 +280,6 @@ class Table extends PureComponent {
         }
       : null
 
-    console.log(extendedToolbar)
-
     return (
       <div className="vtex-table__container">
         {extendedToolbar && (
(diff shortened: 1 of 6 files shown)


react/components/Table/index.js
```diff
@@ -250,12 +255,31 @@ class Table extends PureComponent {
...
+                ...toolbar.inputSearch,
+                onClear: hasBulkActions
+                  ? this.handleInputSearchClearWithBulkAction
+                  : toolbar.inputSearch.onClear,
+              }
+            : null,
+        }
+      : null
+
+    console.log(extendedToolbar)
```
ohmyguigs: spoted
```suggestion
```

react/components/Table/index.js
```diff
@@ -250,12 +255,31 @@ class Table extends PureComponent {
...
+
```
ohmyguigs:
```suggestion
```

react/components/Table/index.js
```diff
@@ -139,6 +139,11 @@ class Table extends PureComponent {
     )
   }
 
+  handleInputSearchClearWithBulkAction = event => {
+    this.handleDeselectAllLines()
+    this.props.toolbar.inputSearch.onClear(event)
```
kevinch: No need to check if the function exists?

ericreis: We are already checking if the inputSearch exists in line 262 (265 to be more specific). The handleInputSearchClearWithBulkActions function in only being used when there are bulk actions and the input search.

Do you think it’s necessary to check if the onCleat function is defined there too?

kevinch: If there's any chance of the function `onClear()` to be called without being defined then yes. Otherwise no.

PR #876, opened 2019-10-03
Hotfix/input search backward compatibility

ohmyguigs:
 - Fixes vtex/styleguide/issues/875

 - input search breaks search 🙃 
[Broken Master](http://pricingqa.myvtex.com/admin/promotions)
[Running workspace](http://inputsearch--pricingqa.myvtex.com/admin/promotions)

 - search for something (it breaks in master, but don't break in the workspace)

#### Types of changes

### DIFF
--- a/react/components/InputSearch/index.js
+++ b/react/components/InputSearch/index.js
@@ -38,9 +38,16 @@ class InputSearch extends Component {
     this.props.onClear && this.props.onClear(event)
   }
 
-  handleSubmit = () => {
+  handleSubmit = event => {
     const { onSubmit, value } = this.props
-    onSubmit && onSubmit(value)
+    onSubmit &&
+      onSubmit({
+        ...event,
+        target: {
+          ...event.target,
+          value,
+        },
+      })
   }
 
   handleHovering = hover => {
@@ -64,7 +71,9 @@ class InputSearch extends Component {
         onBlur={() => this.handleFocus(false)}
         onMouseEnter={() => this.handleHovering(true)}
         onMouseLeave={() => this.handleHovering(false)}
-        onKeyUp={e => e.key === 'Enter' && this.handleSubmit()}
+        onKeyUp={e =>
+          e.key === 'Enter' && this.handleSubmit(new Event('submit'))
+        }
         type="search"
         suffix={
           <div className="flex flex-row items-center">
--- a/react/components/InputSearch/index.js
+++ b/react/components/InputSearch/index.js
@@ -47,6 +47,7 @@ class InputSearch extends Component {
           ...event.target,
           value,
         },
+        preventDefault: event.preventDefault || (() => {}),
       })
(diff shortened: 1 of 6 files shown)


react/components/InputSearch/index.js
```diff
@@ -64,7 +71,9 @@ class InputSearch extends Component {
         onBlur={() => this.handleFocus(false)}
         onMouseEnter={() => this.handleHovering(true)}
         onMouseLeave={() => this.handleHovering(false)}
-        onKeyUp={e => e.key === 'Enter' && this.handleSubmit()}
+        onKeyUp={e =>
+          e.key === 'Enter' && this.handleSubmit(new Event('submit'))
```
kevinch: No risk of the event being `enter` lowercase?

ohmyguigs: In chrome it works like this, tho I did not test in other browsers. :/

alcar: According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values#Whitespace_keys), it seems that it's always `Enter`.

ohmyguigs:
I found another bug in this, will need further testing. I'll continue it on monday ;)

react/components/InputSearch/index.js
```diff
@@ -38,9 +38,17 @@ class InputSearch extends Component {
...
     const { onSubmit, value } = this.props
-    onSubmit && onSubmit(value)
+    onSubmit &&
+      onSubmit({
+        ...event,
+        target: {
+          ...event.target,
+          value,
+        },
+        preventDefault: event.preventDefault || (() => {}),
```
ohmyguigs: @emersonlaurentino @kevinch what do you guys think of this? Since onSubmit SyntheticEvent from react behaviour is to always return the event (which reloads the page in the normal conditions, and bubbles up the DOM), I did this to partially mimic this, so codes where you expect value to be inside `e.target` and event to have a `.preventDefault` don't break.

ohmyguigs: I couldn't think of another solution, if you have some ideas they would be more then welcome.
I also tried creating a new event to pass up (with `new Event('submit')` or `new Event('click')` for example) but they don't have `.preventDefault` when they get to the callback 😔

emersonlaurentino: @ohmyguigs I think you can pass the event as a second part of prop onSubmit and who calls `Input Search` defines event.preventDefault.

ohmyguigs: I just don't wanna break backward compatibility (as before the onSubmit had only the event as param)

emersonlaurentino: I approve your `gambiarra`! I couldn't imagine another way for onClick to take the input value and make it available on target.

Issue #877, opened 2019-10-04
Extra Small Button


matheusps:
##  Overview

Allow `Button` to have a small clickable area to fit on containers that have `2rem` size, such as `Table` cells on `high` density.

## Context
![Screen Shot 2019-10-04 at 10 00 16 (2)](https://user-images.githubusercontent.com/6964311/66209674-42161a00-e68e-11e9-96c4-a8a6e3bf3935.png)

## Examples
Create a new size possibility called `extra-small`, `xsmall` or `tiny`. A good size would be `28px`/`1.75rem`, IMO.

davicosta99:
HI @matheusps, I think #809 will solve this use-case as well.

jonathasbsouza:
@matheusps can we close this issue?

jonathasbsouza:
As we are already using PlainButton, I will close this issue.

Issue #809, opened 2019-08-30
Plain button


davicosta99:
**Describe the need**
We have a tertiary button with props to remove padding on its sides. My proposal is that there should be a plain button that also removes padding on the top and bottom, besides not being all caps.

You can see on the image below two instances were this button variation would be used.

![tertiary](https://user-images.githubusercontent.com/8797476/64045700-5a02f700-cb40-11e9-93db-a184c5f9c88b.png)

**Expected behavior**
It should behave similar to the tertiary button but without it's current footprint and all caps appearance.

Figma: https://www.figma.com/file/LoHnFQmA5FJvapXXWYGhQN/Button?node-id=1%3A45
If you can't access it, talk with me and I'll give you permision.

For reference, see also Polaris' plain button: https://polaris.shopify.com/components/actions/button#navigation

**Screenshots**
![image](https://user-images.githubusercontent.com/8797476/64049526-7d7f6f00-cb4b-11e9-9fd8-7bfa8dca544c.png)

ohmyguigs:
close in https://github.com/vtex/styleguide/pull/886

PR #886, opened 2019-10-11
Feature/button plain

augustocb:

Creates the `ButtonPlain` component.

We decided to create a new button component to use in specific cases, to avoid using `Button` with `collapse` props. It must have its own style and behavior and should not depend on the `Button` component.

You can find more details in these issues:
- https://github.com/vtex/styleguide/issues/809
- https://github.com/vtex/styleguide/issues/877

run `yarn styleguide` and go to http://localhost:6060/#/Components/Forms/ButtonPlain

Screen Shot 2019-10-11 at 14 41 55

#### Types of changes


react/components/ButtonPlain/index.js
```diff
@@ -0,0 +1,248 @@
...
+
+    let classes = 'vtex-button bw1 ba fw5 v-mid relative pa0 br2 bn '
+    let labelClasses = 'flex items-center justify-center h-100 pv1 ttn '
+    let loaderSize = 15
+
+    classes += horizontalCompensation
+    labelClasses += `ph${horizontalPadding} `
+
+    switch (size) {
+      case 'small':
```
matheusps: The `size` should be a constant like:
```
SIZE: {
  SMALL: 'small',
  LARGE: 'large'
  REGULAR: 'regular'
}
```

react/components/ButtonPlain/index.js
```diff
@@ -0,0 +1,248 @@
...
+      labelClasses += 'nowrap '
+    }
+
+    if (iconOnly) {
+      classes += 'icon-button dib '
+    }
+
+    switch (variation) {
+      default:
+      case 'plain': {
```
matheusps: Same as `size`:
```
VARIATION: {
  PLAIN: 'plain',
  INVERTED: 'inverted',
  DANGER: 'danger',
  PRIMARY: 'primary'
}
```

react/components/ButtonPlain/index.js
```diff
@@ -0,0 +1,248 @@
...
+      download,
+      noWrap,
+    } = this.props
+
+    const disabled = this.props.disabled || isLoading
+    const iconOnly = icon || this.props.iconOnly
+    const horizontalPadding = iconOnly ? 1 : 2
+    const horizontalCompensation = `nr${horizontalPadding} nl${horizontalPadding} `
+
+    let classes = 'vtex-button bw1 ba fw5 v-mid relative pa0 br2 bn '
```
matheusps: I'm not a fan of attaching classes latter on for two main reasons:
1. The code becomes spaghetti. Meaning that I need to scroll (up and down) to figure out how the `classes` final value would be.
2. If someone forgot any space at strings tail... Boom 💣! We got a bug in our hands.

Since there is a lot of constraints, using template strings would not be any better (will become a total mess). For that use case, I think `classnames` is very useful.

react/ButtonPlain.js
```diff
@@ -0,0 +1,3 @@
+import ButtonPlain from './components/ButtonPlain/index'
```
matheusps: You can reduce this document by:
```suggestion
export {default}  from './components/ButtonPlain/index'
```

react/components/ButtonPlain/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+      style={style}
+      // Button-mode exclusive props
+      type={href ? undefined : (props.type as 'button' | 'submit' | 'reset')}
+      // Link-mode exclusive props
+      {...(href && linkModeProps)}>
+      {isLoading ? (
+        <Fragment>
+          <span className="top-0 left-0 w-100 h-100 absolute flex justify-center items-center">
+            <Spinner
+              secondary={variation === 'primary' || variation === 'danger'}
```
matheusps:
```suggestion
              secondary={variation === VARIATION.PRIMARY || variation === VARIATION.DANGER}
```

matheusps: And add `PRIMARY` to the constants 🙂

react/components/ButtonPlain/index.tsx
```diff
@@ -0,0 +1,224 @@
...
+      data-testid={props.testId}
+      autoFocus={iconOnly ? undefined : props.autoFocus}
+      disabled={iconOnly ? undefined : props.disabled}
+      name={iconOnly ? undefined : props.name}
+      value={iconOnly ? undefined : props.value}
+      tabIndex={0}
+      className={classes}
+      href={href}
+      onClick={handleClick}
+      onMouseEnter={props.onMouseEnter}
```
Klynger: I think it would be nice if you get `const {  ...rest } = props` and pass it here, it would make it easier to pass some prop that we missed to add and you wouldn't have to pass all these eventlisteners

Klynger:
Another general comment, why `ButtonPlain`? This type of button is usually called `FlatButton`

davicosta99:
@klynger, I’m not sure this would be an appropriate option. Reading about them it seems that “flat” refers to the elevation, and none of our buttons are elevated. Plain, on the other hand, refers to “Plain” text. We saw this name being applied at Shopify Polaris and thought it adequately referenced what differentiated the component.

react/components/ButtonPlain/index.tsx
kevinch: Yes we will do that in the next iteration of the Styleguide, ONDA. But for now we want to limit the abilities of each component deliberately.

Klynger:
> @Klynger, I’m not sure this would be an appropriate option. Reading about them it seems that “flat” refers to the elevation, and none of our buttons are elevated. Plain, on the other hand, refers to “Plain” text. We saw this name being applied at Shopify Polaris and thought it adequately referenced what differentiated the component.

Hmmm, make sense.

react/components/ButtonPlain/index.tsx
marcoskwkm: Oh, nice catch. There's no variation `primary` for this component, this unintentionally remained when we wrote this code based on the `Button` component.

react/ButtonPlain.ts
```diff
@@ -0,0 +1 @@
+export { default } from './components/ButtonPlain/index'
```
Klynger:
```suggestion
export { default } from './components/ButtonPlain'
```

PR #888, opened 2019-10-11
New InputButton component

augustocb:

Creates the InputButton component.

It improves the experience of using a one-input-form.

Details in Figma: https://www.figma.com/file/2t4pmT0LUkPYbohaSxxJ7k/03-InputButton?node-id=0%3A1

Run `yarn styleguide` and go to http://localhost:6060/#/Components/Forms/InputButton.

Screen Shot 2019-10-11 at 19 46 10

#### Types of changes

### DIFF
--- a/react/components/Input/index.js
+++ b/react/components/Input/index.js
@@ -85,7 +85,7 @@ class Input extends Component {
       prefix,
       suffix: suffixProp,
       suffixIcon,
-      submitButton,
+      button,
       groupBottom,
       disabled,
       readOnly,
@@ -246,13 +246,14 @@ class Input extends Component {
               {suffix}
             </span>
           )}
-          {submitButton && size !== 'small' && (
+          {button && size !== 'small' && (
             <span className="flex items-center mr1">
               <Button
-                variation="secondary"
+                disabled={disabled}
                 size={size === 'large' ? 'regular' : 'small'}
-                type="submit">
-                {submitButton}
+                type="submit"
+                variation="secondary">
+                {button}
               </Button>
             </span>
           )}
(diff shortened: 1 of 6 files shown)


ericreis:
I am not sure if this should be solved with a prop or a different component (**InputSubmit**) like other use cases (InputSearch,  InputCurrency, InputPassword, etc.)

augustocb:
@ericreis yes, it should and it has been designed this way!

react/components/Input/index.js
```diff
@@ -1,6 +1,8 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 
+import Button from '../Button'
```
emersonlaurentino: I think you should import the button into the new component and pass it as `suffix` to Input

emersonlaurentino:
I believe this component is not necessary as it is possible to create its behavior with the current input

```jsx
  <Input
    placeholder="Placeholder"
    suffixClasses="flex items-center mr1"
    small="small"
    suffix={
      <Button 
        size="small"
        type="submit"
        variation="secondary"
      >
        Submit
      </Button>
    }
  />
```

cmdalbem:
I tend to agree with Emerson's point. We could add an usage example in the Input component showing how to accomplish this along with the appropriate usage guidelines for this scenario.

omninando:
@emersonlaurentino @cmdalbem they are not possible due to wrong spacing between input and button.

Screen Shot 2019-10-23 at 11 14 03

emersonlaurentino:
@omninando you can add a negative margin to right and it's settled. It's not the most beautiful way to solve it, but I think it's better than creating a new component just for that.

This is a problem that we might have anticipated in `suffix` creation since the "Calculate" button is an input suffix.

kevinch:
Disagree with the negative margin hack. If it's the only solution then we should publish this new component.

ericreis:
Why doesn't it follow the **InputSearch** behavior? Since it's the same idea (an input with a button) I think we should try our best to maintain the whole experience.

cc: @augustocb @cmdalbem 

image

augustocb:
@ericreis don't worry! we are mapping and redesigning all components related to input with buttons.

ericreis:
I reverted these changes since they were merged into master in an unusual way.

cc @augustocb (just to document why I reverted)

PR #891, opened 2019-10-15
Allow grouping conditions subject

ericreis:
Allow developers to group their conditions subjects.

When we had too many subjects it was really hard to find the wanted option. Sometimes we had to use a part of the subject in the verb to overcome this problem (see the image below for an example).

image

[Running workspace](http://10.1.14.209:6060/)

image


cmdalbem:
LGTM 👍
Would be SUPER cool to add in the documentation of the Select component how to use groups.

PR #902, opened 2019-10-22
Move ID attr from div to select element in Dropdown component

augustocb:

It moves `id` attribute from an component's `div` to the `select` element. 

We use this attribute to improve test automation, checking the value of form's fields.

`yarn styleguide`

#### Types of changes

### DIFF
--- a/react/components/Dropdown/index.js
+++ b/react/components/Dropdown/index.js
@@ -180,7 +180,7 @@ class Dropdown extends Component {
         <label>
           {label && <span className={labelClasses}>{label}</span>}
           <div className={containerClasses} style={containerStyle}>
-            <div id={id} className={`vtex-dropdown__button ${classes}`}>
+            <div className={`vtex-dropdown__button ${classes}`}>
               <div className={`flex ${isInline ? '' : 'h-100'}`}>
                 <div
                   className={`vtex-dropdown__caption flex-auto tl truncate ${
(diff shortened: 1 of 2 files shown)


emersonlaurentino:
Hey, @augustocb if the main reason is test. So use prop `testId`, as it is exactly for that. It is a `data-testid` attribute on html element, so you can create tests with *react-testing-library* using `getById` function or create e2e tests with cypress using the function: `cy.get('[data-testid=VALUE_OF_TEST_ID_PROP]')`.

kevinch:
+1 with @emersonlaurentino's suggestion. Plus it would be a breaking change to move the `id`, some apps might use it in some special ways.

react/components/Dropdown/index.js
```diff
@@ -180,7 +180,7 @@ class Dropdown extends Component {
         <label>
           {label && <span className={labelClasses}>{label}</span>}
           <div className={containerClasses} style={containerStyle}>
-            <div id={id} className={`vtex-dropdown__button ${classes}`}>
+            <div className={`vtex-dropdown__button ${classes}`}>
```
kevinch: Revert `id` and add `testid` to the select element. Will be a minor instead of a breaking change.

augustocb:
@omninando could you tell us about your idea of using `id` instead of `testId`?

augustocb:
It already has a `data-testid` attribute. so we are going to use it for a while.

PR #908, opened 2019-10-24
Add staticHeight prop on Modal

mariana-mendes:
as the title says. 

The modal height depends on modal content. 

modalerrado

I needed to use a modal with a fix height. 

modalcerto

[Running workspace](https://minhawks--storecomponents.myvtex.com/admin/app/categories)

If you use the prop `staticHeight`, the max height and min height will be the same. 
And if you don't use the prop, the default behavior is the modal height depends on modal content. 

Clone the repository, checkout to this branch and run yarn && yarn styleguide.

#### Types of changes

### DIFF
--- a/react/components/Modal/index.js
+++ b/react/components/Modal/index.js
@@ -59,6 +59,7 @@ class Modal extends PureComponent {
       showBottomBarBorder,
       onCloseTransitionFinish,
       container,
+      staticHeight,
     } = this.props
     const { shadowBottom, shadowTop } = this.state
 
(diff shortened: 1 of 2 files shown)


react/components/Modal/index.js
```diff
@@ -169,6 +172,8 @@ Modal.propTypes = {
   showTopBar: PropTypes.bool,
   /** Event fired when the closing transition is finished */
   onCloseTransitionFinish: PropTypes.func,
+  /** Delimits modal height when a static height is needed */
+  staticHeight: PropTypes.string,
```
Klynger:
```suggestion
  staticHeight: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
```

PR #912, opened 2019-10-25
Feature/disable numeric stepper

ohmyguigs:

- adds disable prop in NumericStepper

 - pls

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [9.90.4] - 2019-10-25
+
 ### Added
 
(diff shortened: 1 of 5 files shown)


Klynger:
> Maybe `disabled` should be really `disabled` and `readonly` could do this work because semantically you should add `readonly` fields to a form, but `disabled` fields you shouldn't.

You can check this information in the description for disabled: https://developer.mozilla.org/pt-BR/docs/Web/HTML/Element/input

ohmyguigs:
@Klynger I didn't understand. Should I change the prop to `readOnly` and disable the buttons with readOnly?

Klynger:
> @Klynger I didn't understand. Should I change the prop to `readOnly` and disable the buttons with readOnly?

Yeah, since the components you can change the data with the button and they have no action if you try to click it, it should be disabled when the input is readonly

PR #913, opened 2019-10-28
Increase TableV2 performance

matheusps:

🔥 Remove global contexts to improve unnecessary rerenders.
🔥 Avoid tricky children checking and rendering on the `DataTable` component.
🎣 New `useTableMeasures` hook to handle table sizes and current density. This hooks only cares about `density` and `size` making it more flexible and usable to other components, such as`TableTree`.
🎣 New `useTableVisibility` hook that handles columns visibility. This behavior was attached to `useTableState` which is a bad thing because even if the user does not desire to `show/hide` columns it would still be there.
➕ Export `selectedDensity` on `cellRender` to allow users to handle the current density on custom renderers.
💅 Fix toolbar alignment.
💅 Fix `loading` and `emptyState` positions
💣 `useTableState` is now deprecated since it's behaviors are split into `useTableMeasures` and `useTableVisibility`.
🛠 Remove global type declarations.
🛠 Remove unused constants.

[Running workspace](https://matheusps--storecomponents.myvtex.com/admin/collections/), on `admin-collections`

Clone the repo and `git checkout feature/exp-table-performance-boost && yarn && yarn start`

For a short example, the following gifs show the table tree rerenders (represented in green) on toggle the children's visibility.

Before: The entire table is rerendered on any toggle.
before 2019-10-29 12_47_29

After: The table stays static and do not change.
after 2019-10-29 12_48_52
#### Types of changes


react/components/EXPERIMENTAL_Table/index.tsx
```diff
@@ -1,77 +1,96 @@
...
   return (
-    <TableContext.Provider value={{ ...state, ...bulk, ...props, unicityKey }}>
-      <BulkContext.Provider value={bulk}>
-        <TableContainer>
-          {children}
+    <div
+      style={{ minHeight: measures.tableHeight }}
+      id={NAMESPACES.CONTAINER}
+      className="flex flex-column">
+      {children}
```
analuizamtg: indentation?

react/components/EXPERIMENTAL_Table/Toolbar/PopoverMenu.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  const handleOutsideClick = (e: Event) =>
+    buttonRef &&
+    buttonRef.current &&
+    e.target instanceof Node &&
+    !buttonRef.current.contains(e.target) &&
+    setBoxVisible(false)
+
+  useLayoutEffect(() => {
+    isBoxVisible && document.addEventListener('mousedown', handleOutsideClick)
```
estacioneto: I think that when you don't need to return a value, you can use an `if` statement instead of `condition && method()`. It makes more sense when it's an attribution or it's inside the render function.

Klynger:
Workspace is not working (404)

react/components/EXPERIMENTAL_Table/hooks/useTableBulkActions.tsx
```diff
@@ -129,7 +130,20 @@ const useTableBulkActions = ({
...
   )
 
+  const isRowSelected = useCallback(
+    (row: unknown) => {
+      return (
+        bulkState &&
+        bulkState.selectedRows.some(
+          selectedRow => selectedRow[unicityKey] === row[unicityKey]
+        )
+      )
```
estacioneto: Here. It makes sense to use the `condition && method()` because it's returning to someone who will actually use it's value

react/components/EXPERIMENTAL_Table/hooks/useTableVisibility.ts
```diff
@@ -32,13 +29,19 @@ const useHiddenColumns = (columns: Array<Column>) => {
...
   }
 }
 
-export default useHiddenColumns
+export type VisibilityData = {
+  columns: Array<Column>
+}
+
+function getHiddenColumns(columns: Array<Column>): Array<string> {
+  return columns.filter(col => col.hidden).map(col => col.id)
```
estacioneto: I'm curious... Can we use [flatMap](https://developer.mozilla.org/pt-BR/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap)? It would be awesome to use it here!

matheusps:
Fixed @Klynger, just select one collection and check the table below

react/components/EXPERIMENTAL_Table/hooks/useTableVisibility.ts
matheusps: Done

PR #914, opened 2019-10-28
Add Line Chart

fanny:
Add line chart to `styleguide` 

`yarn install` && `yarn start`

#### Types of changes

### DIFF
--- a/react/components/Charts/LineChart/README.md
+++ b/react/components/Charts/LineChart/README.md
@@ -1,3 +1,24 @@
+#### The line chart shows the data as a data points connected by a line. They are useful to analyze changes over the time, comparisons, and trends.
+
+### 👍 Dos
+- *Coming Soon*
+
+### 👎 Don'ts
+- *Coming Soon*
+
+#### Axis
+- *Coming soon*
+
+#### Grid
+- *Coming soon*
+
+#### Datakeys
+- *Coming soon*
+
+#### XaxisKey
+- *Coming soon*
+
+
 ```js
     const data = [
         {
@@ -33,10 +54,10 @@
             tickLine: false
         },
         grid: {
-            horizontal: true,
+            horizontal: false,
             vertical: false,
         }
-    }
+    };
     <LineChart
         data={data}
         dataKeys={keys}
--- a/react/components/Charts/LineChart/README.md
+++ b/react/components/Charts/LineChart/README.md
@@ -1,5 +1,27 @@
 #### The line chart shows the data as a data points connected by a line. They are useful to analyze changes over the time, comparisons, and trends.
 
+```js
+  const sampleData = require('./sampleData').default;
+  const keys = ['customers', 'orders', 'totalSpent'];
+  const mapper = {
+    'customers': 'Customers',
+    'orders': 'Orders',
+    'totalSpent': 'Total Spent'
+  }
+  const formatter = (value, name) => {
+    if(name == 'totalSpent')
+        value = `$${value}`
+    return [value, mapper[name]]
+
+  }
+  <LineChart
+    data={sampleData}
+    dataKeys={keys}
+    xAxisKey='hour'
+    formatter={formatter}
+  />
+```
+
 ### 👍 Dos
 - *Coming Soon*
 
@@ -7,7 +29,7 @@
 - *Coming Soon*
 
 # Line chart schema
-The schema prop defines the style of the chart. This should be given as an object with styles defined for `axis`, `grid`, `container`, according to your needs. 
+The schema prop defines the style of the chart. This should be given as an object with styles defined for `axis`, `grid`, `container`, according to your needs.  
 
 ```js noeditor static
 const schema = {
@@ -42,6 +64,7 @@ The Grid property is responsible to show a grid inside the chart.
 ### Formating values on the tooltip
 
 The formatter prop takes a `function` which will be used to render each content of tooltip. If you want to customize the content of your tooltip, make sure that your function follows this pattern:
+
 ```jsx noeditor static
 const mapper = {
   'customers': 'Customers',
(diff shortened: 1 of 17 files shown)


react/components/Charts/LineChart/README.md
```diff
@@ -0,0 +1,46 @@
...
+        },
+        grid: {
+            horizontal: true,
+            vertical: false,
+        }
+    }
+    <LineChart
+        data={data}
+        dataKeys={keys}
+        xAxisKey={'name'}
```
matheusps:
```suggestion
        xAxisKey="name"
```

react/components/Charts/LineChart/constants.ts
```diff
@@ -0,0 +1,15 @@
...
+    '#479A5E'
+]
+type DefaultLineProps = Pick<LineProps, 'type' | 'strokeWidth' | 'dot'>
+const defaultProps: DefaultLineProps = {
+    type: 'monotone',
+    strokeWidth: 3,
+    dot: false,
+}
+
+export { colors, defaultProps }
```
matheusps: Make your exports explicit on the constant declaration looks better than a single export at the file's end. (IMO)

react/components/Charts/LineChart/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+    YAxis,
+    Tooltip,
+    CartesianGrid,
+    ResponsiveContainer,
+} from 'recharts'
+import { colors, defaultProps } from './constants'
+import useChart from '../hooks/useChart'
+
+interface Props {
+    data: any,
```
matheusps:
```suggestion
    data: object[],
```
maybe ?

react/components/Charts/LineChart/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+    dataKeys: string[],
+    xAxisKey: string,
+    yAxisKey?: string,
+    schema: ChartProps
+}
+
+const renderLines = (keys: string[]) => {
+    return keys.map((key, i) =>
```
matheusps: You do not need this `return`

react/components/Charts/LineChart/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+    yAxisKey?: string,
+    schema: ChartProps
+}
+
+const renderLines = (keys: string[]) => {
+    return keys.map((key, i) => 
+      <Line
+        key={key}
+        dataKey={key}
+        stroke={colors[i]}
```
matheusps: If you have more keys than colors this will cause an error.

react/components/Charts/commonProps.ts
```diff
@@ -0,0 +1,16 @@
+const commonDefaultProps = {
```
matheusps: Why it's not on the `constants` file?

react/components/Charts/hooks/useChart.ts
```diff
@@ -0,0 +1,8 @@
+import { commonDefaultProps } from '../commonProps'
+
+const useChart = (userProps: ChartProps)  => {
```
matheusps: This hook does not seem that useful, you could just import the `commonDefaultProps` on the desired component. And you would do something like:
```tsx
// Component File
import { commonDefaultProps } from 'constants'

const Component: FC<Props> = props => { ... }

Component.defaultProps = {
  ...commonDefaultProps
}

interface Props  extends ChartProps {
  // otherProps
}
```

react/components/Charts/LineChart/constants.ts
angoncal: I agree with Matheus...
You can do like `export const defaultProps: DefaultLineProps = {
    type: 'monotone',
    strokeWidth: 3,
    dot: false,
}`

react/components/Charts/LineChart/index.tsx
angoncal: `const renderLines = (keys: string[]) => (keys.map((key, i) => ....))`

react/components/Charts/LineChart/constants.ts
```diff
@@ -0,0 +1,15 @@
+import { LineProps } from 'recharts'
+const colors = [
+    '#EE7850',
+    '#4CAABF',
+    '#9F50B6',
+    '#479A5E'
+]
+type DefaultLineProps = Pick<LineProps, 'type' | 'strokeWidth' | 'dot'>
+const defaultProps: DefaultLineProps = {
+    type: 'monotone',
```
angoncal: I think you should use the 'linear' type

react/components/Charts/LineChart/index.tsx
fanny: Actually, i'm having troubles with proptypes, because i define data with a type in typescript, so when i try to put `object` the proptypes complain of the shape. Do you have some suggestion?

react/components/Charts/LineChart/constants.ts
fanny: Following the figma, it seems that is `monotone`, check it out: https://www.figma.com/file/G2jbai2bVLw1u7JOSbw6R9/02-Charts

react/components/Charts/LineChart/index.tsx
fanny: Only four lines are allowed, with more, the chart would be confused, but i will think in a better approach

react/components/Charts/hooks/useChart.ts
fanny: Because this object sets some props that aren't editables on the component, and used in each `recharts` component. But i agree with you that the hook is not useful, i can import only the object, and merge with the props defined by the user.

react/components/Charts/commonProps.ts
fanny: Because this props are common for many graphs, not only for line charts, that contains the `constants` file.

react/components/Charts/LineChart/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+        <LineChartBase data={data}>
+            <XAxis dataKey={xAxisKey} {...configs.axis} />
+            <YAxis {...configs.axis}/>
+            <CartesianGrid
+                horizontal={configs.grid.horizontal}
+                vertical={configs.grid.vertical}
+            />
+            <Tooltip formatter={formatter} {...tooltipProps}/>
+            {zipWith(renderLines, dataKeys, colors)}
+        </LineChartBase>
```
emersonlaurentino: change the spaces from 4 to 2, pls

fanny: Oh sorry, fixed! :laughing:

react/components/Charts/LineChart/index.tsx
```diff
@@ -0,0 +1,76 @@
+import React, { FC } from 'react'
+import { zipWith } from 'ramda'
```
kevinch: I would not import ramda just for one function. The project uses lodash now.

fanny: Oh, okay. i will update

react/components/Charts/LineChart/constants.ts
```diff
@@ -0,0 +1,23 @@
+import { LineProps } from 'recharts'
+
+export const colors = [
+  '#EE7850',
+  '#4CAABF',
+  '#9F50B6',
+  '#479A5E',
+  '#0000FF' //sample
```
kevinch: Useful comment?

fanny: Opss

PR #915, opened 2019-10-28
Add optional onRowHover prop to table

anitavincent:

To add an optional prop `onRowHover` to table

To allow devs to take action when some line is hovered, like specific changes to some cells in the hovered row.

The motivation comes from our team designer wanting some buttons to pop up only when the row is hovered.

https://anita--partnerchallenge26.myvtex.com/admin/received-skus/pending/ has an example of a table using onRowHover
click on the "Approved" tab and see another table, this time without `onRowHover`, still working properly.

#### Types of changes

### DIFF
--- a/react/components/Table/SimpleTable.js
+++ b/react/components/Table/SimpleTable.js
@@ -56,7 +56,7 @@ class SimpleTable extends Component {
   }
 
   handleRowHover = rowIndex => {
-    const { onRowClick } = this.props
+    const { onRowClick, onRowHover } = this.props
     const { isLineActionsHovered } = this.state
     if (onRowClick && !isLineActionsHovered) {
       this.setState({
@@ -67,6 +67,8 @@ class SimpleTable extends Component {
         hoverRowIndex: -1,
       })
     }
+
+    onRowHover(rowIndex)
   }
 
   calculateColWidth = (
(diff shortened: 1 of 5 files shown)


react/components/Table/SimpleTable.js
```diff
@@ -56,7 +56,7 @@ class SimpleTable extends Component {
   }
 
   handleRowHover = rowIndex => {
-    const { onRowClick } = this.props
+    const { onRowClick, onRowHover = () => {} } = this.props
```
omninando: shouldn't this be a default prop instead of a default paramater?

react/components/Table/index.js
```diff
@@ -439,8 +443,8 @@ Table.propTypes = {
     texts: PropTypes.shape({
       secondaryActionsLabel: PropTypes.string.isRequired,
       rowsSelected: PropTypes.func.isRequired,
-      selectAll: PropTypes.string.isRequired,
-      allRowsSelected: PropTypes.func.isRequired,
```
felippenardi: Do you need to remove this `isRequired`?

anitavincent: This was actually a sneaky bug fix 😅  I made them optional in another PR but didn't change it here. Now it keeps alerting `selectAll is required!!!!!1!!one`. But it isn't

react/components/Table/SimpleTable.js
anitavincent: Yes! Fixed it

react/components/Table/index.js
felippenardi: Got you :)

I believe you could leave this change off this PR, and git will add those back when you merge or rebase the PR to master.

PR #922, opened 2019-11-01
Provide updateCellMeasurements() method for cell height refresh

arturpimentel:
Add an `updateCellMeasurements()` callback to the provided props in `skuCellRenderer`.

While the table height is being updated whenever there is a different *sum* of rows heights, sometimes the individual cell height is not being updated in the cell measurement cache, causing said sum to convey the wrong height measurement for the table.

Since this depends on some obscure logic for the CellMeasurer cache that maintains the cell previous measurements, one way of refreshing those measurements is to provide the `measure()` render prop provided by the `CellMeasurer`. That way one can programatically remeasure an individual cell and update the entire row.

This is similar to the `react-virtualized` recommended approach for cells that contain images, [which can be seen here](https://github.com/bvaughn/react-virtualized/blob/master/docs/CellMeasurer.md#using-cellmeasurer-with-images)

To see the problem:
1. [Visit this link](https://measuretest--partnerchallenge26.myvtex.com/admin/received-skus/pending)
2. Check a SKU
3. Click on the `Link to existing SKU` button
4. When the modal shows up, you'll see that there is an input.
5. Write an random string of letters only (e.g. `sdkfjas`)
6. Confirm by clicking on `Link to SKU`
7. You should see that there is an error message at the SKU line, but it touches the bottom line of the row (padding is hidden)

To see the working fix:
1. [Visit this link](https://artur--partnerchallenge26.myvtex.com/admin/received-skus/pending)
2. Check a SKU
3. Click on the `Link to existing SKU` button
4. When the modal shows up, you'll see that there is an input.
5. Write an random string of letters only (e.g. `sdkfjas`)
6. Confirm by clicking on `Link to SKU`
7. You should see that there is an error message at the SKU line, but now the padding is showing up

Before:
image

After:
image

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [9.96.0] - 2019-11-07
+
 ### Added
 
 - `updateCellMeasurements` function in **Table** to be used in `cellRenderer` in order to update cell size programatically
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 
-- to **Table** a `updateCellMeasurements` function to be used in `cellRenderer` in order to update cell size programatically
+- `updateCellMeasurements` function in **Table** to be used in `cellRenderer` in order to update cell size programatically
 
 ## [9.95.0] - 2019-11-07
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- to **Table** a `updateCellMeasurements` function to be used in `cellRenderer` in order to update cell size programatically
+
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- to **Table** a `updateCellMeasurements` function to be used in `cellRenderer` in order to update cell size programatically
```
kevinch:
```suggestion
- `updateCellMeasurements` function in **Table** to be used in `cellRenderer` in order to update cell size programatically
```

react/components/Table/SimpleTable.js
```diff
@@ -359,60 +359,67 @@ class SimpleTable extends Component {
...
+                                  fullWidth,
+                                  fullWidthColWidth
+                                ),
+                                backgroundColor: selectedRowsIndexes.includes(
+                                  rowIndex - 1
+                                )
+                                  ? SELECTED_ROW_BACKGROUND
+                                  : '',
+                              }}
+                              className={`flex items-center w-100 h-100 ph4 bb
```
kevinch: Nitpick: extract this logic outside the render function to keep readability.

```
const classNames = `flex items-center w-100 h-100 ph4 bb ... ${} ...`

return () {
...
className={classNames}
```

PR #926, opened 2019-11-04
Add loading status to InputButton

augustocb:

To allow a loading status to `InputButton` component.

`InputButton` don't have a way to give feedback when, after submitting the form, something needs to load and the user needs to wait.

http://localhost:6060/#/Components/Forms/InputButton

loding

#### Types of changes

### DIFF
--- a/react/components/Input/index.js
+++ b/react/components/Input/index.js
@@ -126,8 +126,10 @@ class Input extends Component {
       classes += 'code '
     }
 
-    if (disabled || readOnly) {
-      classes += `bg-transparent b--disabled ${disabled ? 'c-disabled' : ''} `
+    if (disabled || readOnly || isLoadingButton) {
+      classes += `bg-transparent b--disabled ${
+        disabled || isLoadingButton ? 'c-disabled' : ''
+      } `
       prefixSuffixGroupClasses += `bg-disabled b--disabled ${
         disabled ? 'c-disabled' : ''
       } `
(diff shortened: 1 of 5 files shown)


react/components/Input/index.js
```diff
@@ -86,6 +86,7 @@ class Input extends Component {
       suffix: suffixProp,
       suffixIcon,
       button,
+      isLoadingButton,
```
kevinch: I'd change to `isLoading` since it's a pattern commonly used in other components.

augustocb: @kevinch I'm afraid of using `isLoading` on the `Input` component and looking like `Input` status instead of a `Button` status.

matheusps: Just `loading` wouldn't be enough? 🤔

kevinch: or `loading` yes. I'll add a note for the next iteration of the Styleguide to set a pattern for this kind o variable name.

emersonlaurentino: i opened a thread about it: https://vtex.slack.com/archives/C1MCRG7CK/p1571774967090800

augustocb: BUT we can think about a loading state for the `Input` component itself as well :). later 😬

kevinch: Sounds like it's a loading state for the whole component since i can see on the gif the input turning "disabled".

kevinch: Sounds like it's a loading state for the whole component since i can see on the gif the input turning "disabled".

augustocb: @kevinch it's just an optical illusion.. the input and the button have the same bg color.

PR #933, opened 2019-11-07
Fix/radio

augustocb:

It removes an extra margin from last Radio of RadioGroup.

http://localhost:6060/#/Components/Forms/RadioGroup

Before:

Screen Shot 2019-11-07 at 16 40 30

After:

Screen Shot 2019-11-07 at 16 40 40

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+- Extra margin from last Radio in the RadioGroup list.
+
 ## [9.94.0] - 2019-11-07
 
 ### Fixed
--- a/react/components/Radio/index.js
+++ b/react/components/Radio/index.js
@@ -21,6 +21,7 @@ class Radio extends PureComponent {
       checked,
       disabled,
       id,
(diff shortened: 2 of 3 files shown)


emersonlaurentino:
A new prop for this is not necessary. You can add a new css class, something like this:

```css
.radioWrapper:last-child {
 margin-bottom: 0px;
}
```

and inside the component you call it this:

```jsx
// ...Radio
  className={classNames(styles.radioWrapper, 'flex items-start mb3 relative', {
    pointer: !disabled,
  })}
// ... Radio
```

This will make sure that only the last child (radio) does not have a `margin-bottom` and has the advantage that you do not have to put a new prop on all children.

augustocb:
@emersonlaurentino I'm not sure this approach (overwriting the margin using css) is the best onde. What you think @omninando @kevinch?

emersonlaurentino:
You are not overwriting css, this is trivial. `last-child` and` first-child` exist for just that.

augustocb:
@emersonlaurentino Yes, I know! But that css rule overwrites the `mb3` rule.

emersonlaurentino:
Yes, because `mb3` doesn't makes no sense to the last child. But I will approve this PR, because the children of `RadioGroup` are structured in a way that css does not understand.

PR #939, opened 2019-11-12
Show EXPERIMENTAL_BulkActions even if no action is set

vitoriaHeliane:
As the title says.

In the `ProductsTable` of the `admin-collections` we need to use the bulkActions without set any action to it, according to the design as you can see bellow. For now, if the main and the other actions are not set, the component does not appear.
image

Run the following command and access the `http://localhost:6060/`:

```
yarn start
```

Or access the [workspace](https://waza--storecomponents.myvtex.com/admin/collections/1103/).

#### Types of changes

### DIFF
--- a/react/components/EXPERIMENTAL_Table/BulkActions.tsx
+++ b/react/components/EXPERIMENTAL_Table/BulkActions.tsx
@@ -25,7 +25,6 @@ const BulkActions: FC<BulkActionsProps> = ({
     bulkState,
     selectAllRows,
     deselectAllRows,
-    hasBulkActions,
     hasPrimaryBulkAction,
     hasSecondaryBulkActions,
   } = data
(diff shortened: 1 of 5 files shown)


react/components/EXPERIMENTAL_Table/BulkActions.tsx
```diff
@@ -51,30 +50,28 @@ const BulkActions: FC<BulkActionsProps> = ({
...
+          <ActionMenu
+            label={texts.secondaryActionsLabel}
+            buttonProps={{ variation: 'secondary', size: 'small' }}
+            options={others.map(el => ({
+              label: el.label,
+              onClick: () => el.onClick(bulkActionsReturnedParameters),
+            }))}
+          />
+        )}
+      </div>
```
estacioneto: I think this change is unnecessary, actually... @matheusps has an opinion about that?

PR #950, opened 2019-11-22
Add page top indicator option to Pagination

arturpimentel:
Add a page top indicator option to the `Pagination` component

Some pages have content that are so large that the reference of which page is being displayed is lost to the immediate view.

To fix this, add an option to the pagination component in which a page indicator is shown above the content in such a way that the paginated nature of a search result is easier to perceive.

You can [visit this link](https://artur--partnerchallenge26.myvtex.com/admin/received-skus/approved/?elements=15&from=0&to=15) to see an example of a table using the proposed feature

image

#### Types of changes

### DIFF
--- a/react/components/Pagination/PageIndicator.js
+++ b/react/components/Pagination/PageIndicator.js
@@ -4,18 +4,17 @@ import PropTypes from 'prop-types'
 function PageIndicator({ currentItemFrom, itemTo, textOf, totalItems }) {
   return (
     <div className="c-muted-2 t-small">
-      {currentItemFrom}
-      {' - '}
-      {itemTo} {textOf} {totalItems}
+      {`${currentItemFrom} - ${itemTo} `}
+      {textOf} {totalItems}
     </div>
   )
 }
 
 PageIndicator.propTypes = {
-  currentItemFrom: PropTypes.number,
-  itemTo: PropTypes.number,
-  textOf: PropTypes.string,
-  totalItems: PropTypes.number,
+  currentItemFrom: PropTypes.number.isRequired,
+  itemTo: PropTypes.number.isRequired,
+  textOf: PropTypes.string.isRequired,
+  totalItems: PropTypes.number.isRequired,
 }
 
 export default PageIndicator
--- a/react/components/Pagination/PageIndicator.js
+++ b/react/components/Pagination/PageIndicator.js
@@ -1,0 +1,21 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+function PageIndicator({ currentItemFrom, itemTo, textOf, totalItems }) {
+  return (
+    <div className="c-muted-2 t-small">
+      {currentItemFrom}
+      {' - '}
+      {itemTo} {textOf} {totalItems}
+    </div>
+  )
+}
+
+PageIndicator.propTypes = {
+  currentItemFrom: PropTypes.number,
+  itemTo: PropTypes.number,
+  textOf: PropTypes.string,
+  totalItems: PropTypes.number,
+}
+
+export default PageIndicator
--- a/react/components/Pagination/PageIndicator.js
+++ b/react/components/Pagination/PageIndicator.js
@@ -1,20 +1,31 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 
-function PageIndicator({ currentItemFrom, itemTo, textOf, totalItems }) {
+function PageIndicator({
+  currentItemFrom,
+  itemTo,
+  textOf,
+  totalItems,
+  itemLabel,
+}) {
   return (
     <div className="c-muted-2 t-small">
       {`${currentItemFrom} - ${itemTo} `}
-      {textOf} {totalItems}
+      {textOf} {totalItems} {itemLabel}
     </div>
   )
+}
+
+PageIndicator.defaultProps = {
+  itemLabel: '',
 }
 
 PageIndicator.propTypes = {
   currentItemFrom: PropTypes.number.isRequired,
   itemTo: PropTypes.number.isRequired,
   textOf: PropTypes.string.isRequired,
-  totalItems: PropTypes.number.isRequired,
+  totalItems: PropTypes.node.isRequired,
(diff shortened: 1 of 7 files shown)


react/components/Pagination/PageIndicator.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+function PageIndicator({ currentItemFrom, itemTo, textOf, totalItems }) {
+  return (
+    <div className="c-muted-2 t-small">
+      {currentItemFrom}
+      {' - '}
+      {itemTo} {textOf} {totalItems}
+    </div>
```
emersonlaurentino: you can use this:
```jsx
    <div className="c-muted-2 t-small">
        {`${currentItemFrom} - ${itemTo} ${textOf} ${totalItems}`}
    </div>
```

react/components/Pagination/PageIndicator.js
```diff
@@ -0,0 +1,21 @@
...
+  )
+}
+
+PageIndicator.propTypes = {
+  currentItemFrom: PropTypes.number,
+  itemTo: PropTypes.number,
+  textOf: PropTypes.string,
+  totalItems: PropTypes.number,
```
emersonlaurentino: all should be `isRequired`

react/components/Pagination/index.js
```diff
@@ -5,6 +5,8 @@ import CaretRight from '../icon/CaretRight'
 import ButtonWithIcon from '../ButtonWithIcon'
 import Dropdown from '../Dropdown'
```
emersonlaurentino: remove line

react/components/Pagination/index.js
```diff
@@ -60,49 +63,64 @@ class Pagination extends PureComponent {
...
-              icon={caretRight}
-              variation="secondary"
-              size="small"
-              disabled={isNextDisabled}
-              onClick={this.handleNextPage}
+        {this.props.children}
+        <div
+          className={`flex flex-row items-center ${
+            rowsOptions ? 'justify-between' : 'justify-end'
+          }`}>
```
emersonlaurentino: why do you not use `classnames` function?

react/components/Table/SwitchablePagination.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
```
emersonlaurentino: remove line

react/components/Table/SwitchablePagination.js
```diff
@@ -0,0 +1,18 @@
...
+function SwitchablePagination({ children, enabled, ...paginationProps }) {
+  if (enabled) {
+    return <Pagination {...paginationProps}>{children}</Pagination>
+  }
+  return children
+}
+
+SwitchablePagination.propTypes = {
+  ...Pagination.propTypes,
+  enabled: PropTypes.bool,
```
emersonlaurentino: since enabled is not required, you use `defaultProps` as `false`, however undefined is a false value.

arturpimentel: Done

react/components/Pagination/index.js
arturpimentel: Done, although there is just one flag so far, for the sake of expandability.

react/components/Pagination/PageIndicator.js
arturpimentel: Done

react/components/Pagination/PageIndicator.js
arturpimentel: Partially done. One can aggregate the `${currentItemFrom} - ${itemTo}` part, but not the others, because in reality `textOf` can be a `<FormattedMessage />`, and implementing your suggestion exactly can render a string like `1 - 10 [object Object] 300` because the template string calls `.toString()` for its parameters.

react/components/Pagination/index.js
arturpimentel: I'll actually keep the blank line, along with line 3. Other components in the Styleguide also organize their imports in the General -> Specific groups separated by blank lines. I believe this adds a little bit of readability.

react/components/Pagination/PageIndicator.js
emersonlaurentino: but `textOf` can never be `<FormattedMessage />` since your PropType expects it to be a `string`.

react/components/Pagination/index.js
emersonlaurentino: all right

react/components/Pagination/PageIndicator.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+function PageIndicator({ currentItemFrom, itemTo, textOf, totalItems }) {
+  return (
+    <div className="c-muted-2 t-small">
+      {`${currentItemFrom} - ${itemTo} `}
```
laerc:
```suggestion
      {`${currentItemFrom} - ${itemTo}`}
```
Does the space here is something that you want?

arturpimentel: Yes, it adds a space between `itemTo` and `textOf`

arturpimentel:
@laerc this text disposition is the current demand [as can be seen here](https://www.figma.com/file/IFY7g2VaBjfOyyxUS2AYJxGG/(m)-receiveds-skus-ds?node-id=1927%3A2080). In the future we may add alignment options like the ones that exist for the Styleguide buttons.

react/components/Pagination/PageIndicator.js
arturpimentel: That's true, but further inspection reveals that `Pagination/index.js` expects `PropTypes.node.isRequired` for `textOf`. If I change this, current behaviour can change for some of our dependants. Which means... we have a choice:
1. Risk causing display bugs for some apps that are using `<FormattedMessage />` and aggregate all props in the same template message as expected strings.
2. Change `textOf` prop to `node` in `PageIndicator.js` and keep things as they are.

PR #953, opened 2019-11-27
Table V2 Documentation (State Handlers and Features)

matheusps:
#### What is the purpose of this pull request?

As the title says.
It also removes unused props.

#### What problem is this solving?

Great components become awesome with great documentation. 💅

#### How should this be manually tested?

Clone the repo and `yarn && yarn start`

#### Types of changes

- [ ] Bug fix (a non-breaking change which fixes an issue)
- [ ] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Requires change to documentation, which has been updated accordingly.

PR #955, opened 2019-11-28
Compound Bulk Actions

matheusps:

♻️Makes bulk actions composable to improve flexibility throughout multiple use cases.
💅The order of components is granted, like on the toolbar.
🎯Meaningful eCommerce examples.
📚The context documentation will be added on the full documentation PR (#953 )
🛠Fixed the default comparator of `useCheckoxesTree`
🛠Fixed error on not passing the `onToggle` function to `useCheckboxesTree`

The `BulkActions` did not have the desired support for conditional rendering that occurs in the result of micro-interactions. Now the internal components render are fully controlled by the developer, but order and the actual look cannot be changed.

clone the repo and `yarn && yarn start`.

No need.

#### Types of changes


react/components/EXPERIMENTAL_Table/BulkActions/Actions.tsx
```diff
@@ -0,0 +1,61 @@
...
+type PrimaryProps = {
+  label: string
+  onClick: () => void
+}
+
+type SecondaryProps = {
+  label: string
+  actions: Array<MenuAction>
+  onActionClick: (el: MenuAction) => void
+}
```
mariana-mendes: I would just change the order of definitions in the file. I think it would be more readable. :thinking: 
(type definitions before functions definitions)

mariana-mendes:
Also, verify `FULL EXAMPLE`, maybe `Decrease` action has a problem.

react/components/EXPERIMENTAL_Table/BulkActions/Actions.tsx
matheusps: I prefer the definitions on the bottom because I like to read directly the component content instead of scrolling through a bunch of type definitions. The full definition doesn't matter most of the time, especially if you just opened the file and did not saw in which context are these types beeing used. `OBS`: If you use VS Code you can preview the type by just hovering it with the mouse

react/components/EXPERIMENTAL_useCheckboxTree/index.tsx
```diff
@@ -60,7 +60,7 @@ export default function useCheckboxTree<T>({
   }, [checkedItems])
 
   const isChecked = useCallback(
-    (item: T) => {
+    (item: T | Tree<T>) => {
       const childs = item[nodesKey] as Array<T>
```
Klynger: Could you fix this miss spelling pls?

PR #966, opened 2019-12-12
Remove key from option

rerissondaniel:
Remove the key prop from the option button.

The key is already defined [here](https://github.com/vtex/styleguide/blob/422a40f7c9715a6bcd45578e18de50f22e7bd0e1/react/components/AutocompleteInput/index.tsx#L160)

The key only needs to be defined in the upper level component

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [9.103.2] - 2019-12-20
+
 ### Removed
 
 - `key` that was used as prop in **AutocompleteInput**.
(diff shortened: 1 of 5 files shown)


react/components/AutocompleteInput/Option.tsx
```diff
@@ -72,7 +70,6 @@ const Option: React.FunctionComponent<
 
   return (
     <button
-      key={key}
```
kevinch: No chance if this breaking anything?

rerissondaniel: No, key should never be read from props, the behavior in that case is not specified (it was just undefined).

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Removed
+
+-Remove `key` that was used as prop.
```
emersonlaurentino:
```suggestion
- `key` that was used as prop in **AutocompleteInput**.
```

PR #967, opened 2019-12-12
Added useTableProportion hook

matheusps:

- Add useTableProportion hook.
- Solve cell bugs.

⚠️ The detailed documentation will be present on #953 

You can consider this hook a shorthand for setting columns width based on a passed ratio array.

Clone the repo and `yarn && yarn start`

This represents the ratio `[3,1]`. This means that the first column will grow three times more than the second.
Screen Shot 2019-12-12 at 10 13 39

If we want to divide equaly, just  set 1 to every col `[1,1]`
Screen Shot 2019-12-12 at 10 14 11

⚠️ The proportions respect the `truncate of rows`. This means that doesn't matter how large one ratio is compared to other ones - the content will **NOT** break by `default`. In the screenshot below you can see the ratio `[100, 1]`. 
Screen Shot 2019-12-12 at 10 15 06

#### Types of changes

### DIFF
--- a/react/components/EXPERIMENTAL_Table/hooks/useTableProportion.ts
+++ b/react/components/EXPERIMENTAL_Table/hooks/useTableProportion.ts
@@ -19,11 +19,10 @@ export default function useTableProportion({ columns, ratio }: ProportionData) {
 
 function calculateWidths(columns: Array<Column>, ratio: Array<number>) {
   const slicedRatio = ratio.slice(0, columns.length)
-
   const sum = (acc: number, num: number) => acc + num
   const ratioSum = slicedRatio.reduce(sum, 0)
-  const minPortion = 100 / ratioSum
-
+  const fullPortion = 100
+  const minPortion = fullPortion / ratioSum
   return slicedRatio.map(value => `${value * minPortion}%`)
 }
 
--- a/react/components/EXPERIMENTAL_Table/hooks/useTableProportion.ts
+++ b/react/components/EXPERIMENTAL_Table/hooks/useTableProportion.ts
@@ -1,0 +1,33 @@
+import { useMemo } from 'react'
+import { Column } from '../index'
+
+export default function useTableProportion({ columns, ratio }: ProportionData) {
+  const calculatedWidths = useMemo(() => calculateWidths(columns, ratio), [
+    columns,
+    ratio,
+  ])
+
+  const sizedColumns = useMemo(
+    () => columns.map((col, i) => ({ ...col, width: calculatedWidths[i] })),
+    [columns, ratio]
+  )
+
+  return {
+    sizedColumns,
+  }
+}
+
+function calculateWidths(columns: Array<Column>, ratio: Array<number>) {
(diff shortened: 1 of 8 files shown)


react/components/EXPERIMENTAL_Table/hooks/useTableProportion.ts
```diff
@@ -0,0 +1,33 @@
...
+  )
+
+  return {
+    sizedColumns,
+  }
+}
+
+function calculateWidths(columns: Array<Column>, ratio: Array<number>) {
+  const slicedRatio = ratio.slice(0, columns.length)
+
```
emersonlaurentino: why this line?

react/components/EXPERIMENTAL_Table/hooks/useTableProportion.ts
```diff
@@ -0,0 +1,33 @@
...
+  const sum = (acc: number, num: number) => acc + num
+  const ratioSum = slicedRatio.reduce(sum, 0)
+  const minPortion = 100 / ratioSum
+
+  return slicedRatio.map(value => `${value * minPortion}%`)
```
emersonlaurentino: Why can not return number?

react/components/EXPERIMENTAL_Table/hooks/useTableProportion.ts
```diff
@@ -0,0 +1,33 @@
...
+    sizedColumns,
+  }
+}
+
+function calculateWidths(columns: Array<Column>, ratio: Array<number>) {
+  const slicedRatio = ratio.slice(0, columns.length)
+
+  const sum = (acc: number, num: number) => acc + num
+  const ratioSum = slicedRatio.reduce(sum, 0)
+  const minPortion = 100 / ratioSum
```
mariana-mendes: 100. could be a const with a meaningful name

matheusps: you're right

react/components/EXPERIMENTAL_Table/hooks/useTableProportion.ts
matheusps: This function is meant to return the array of percentages for each column. I represented a string so that I do not need to do it later since the manipulation is over.

react/components/EXPERIMENTAL_Table/hooks/useTableProportion.ts
matheusps: done

react/components/EXPERIMENTAL_Table/hooks/useTableProportion.ts
matheusps: oops, done

react/components/EXPERIMENTAL_Table/hooks/useTableProportion.ts
emersonlaurentino: all right

PR #988, opened 2020-01-10
Fix EXPERIMENTAL_Table toolbar responsiveness on mobile

thayannevls:
Make toolbar items wrap on small screens, avoiding items to overflow.

When the toolbar has to many items it's causing the toolbar to overflow the table.

`yarn && yarn start`

**Before**
Screenshot from 2020-01-10 13-19-20

**After**
Screenshot from 2020-01-10 13-20-32
#### Types of changes

### DIFF
--- a/react/components/EXPERIMENTAL_Table/Toolbar/InputSearch.tsx
+++ b/react/components/EXPERIMENTAL_Table/Toolbar/InputSearch.tsx
@@ -1,5 +1,4 @@
 import React, { FC, FormEvent, InputHTMLAttributes } from 'react'
-import csx from 'classnames'
 
 import Input from '../../InputSearch/index'
 import { NAMESPACES, ORDER_CLASSNAMES } from '../constants'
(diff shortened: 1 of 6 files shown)


react/components/EXPERIMENTAL_Table/Toolbar/InputSearch.tsx
```diff
@@ -12,7 +11,7 @@ const InputSearch: FC<InputSearchProps> = ({ onSubmit, ...inputProps }) => {
   return (
     <form
       id={NAMESPACES.TOOLBAR.INPUT_SEARCH}
-      className={csx(ORDER_CLASSNAMES.TOOLBAR_CHILD.INPUT, 'w-40')}
+      className={ORDER_CLASSNAMES.TOOLBAR_CHILD.INPUT}
```
matheusps:
```suggestion
      className={csx(ORDER_CLASSNAMES.TOOLBAR_CHILD.INPUT, 'w-40-ns')
```
maybe ?

matheusps: You could also add `w-100-s` to make it more appealing.

thayannevls: It works well only if put `flex-wrap`only in small screens, because of this:

Screenshot from 2020-01-10 14-04-17

But I'm not so sure if it's good to wrap only on small screens.

matheusps: Oh, I see. I think is up to the design team to decide if it's okay.

PR #994, opened 2020-01-14
EXPERIMENTAL_Table motion

matheusps:
#### 🔥Added table motions

♻️Extensible and reusable.
🚫No external libraries. 
🖥Respects the user accessibility motion definitions.

[Running workspace](https://tablemotion--cosmetics1.myvtex.com/admin/app/collections)

Clone the repo and `yarn && yarn start`. You should test for reduced and not reduced motion.

On density change
density

On bulk actions toggle
bulk

#### Types of changes


react/components/EXPERIMENTAL_TableTree/Tree/Node.tsx
```diff
@@ -8,6 +8,17 @@ import CollapseToggle from './CollapseToggle'
 import { Checkboxes } from '../../EXPERIMENTAL_useCheckboxTree/types'
 import { Column } from '../../EXPERIMENTAL_Table'
 import { Density } from '../../EXPERIMENTAL_Table/hooks/useTableMeasures'
+import useTableMotion from '../../EXPERIMENTAL_Table/hooks/useTableMotion'
+
+const TRANSITIONS = [
```
kevinch: Are those the same as `BulkActions/index.tsx`?

matheusps: It was the same as rows, nice catch!

react/components/EXPERIMENTAL_Table/hooks/useTableMotion.ts
```diff
@@ -0,0 +1,79 @@
...
+    }
+    mql.addEventListener('change', throttle(handleChange, 200))
+    return () => {
+      mql.removeEventListener('change', handleChange)
+    }
+  }, [])
+  return reduced
+}
+
+type transition = {
```
rerissondaniel: I always like to use uppercase for types, but I know that is not enforced

matheusps: I agree with you in this case. Nice catch, btw. When my type expresses a callback or function I use camelCase though

PR #997, opened 2020-01-16
Add missing handle to the label of the Button

Klynger:

Add a missing handle to the label of the Button

[Running workspace](https://buttonhandle--storecomponents.myvtex.com/)

The agencies want to style the label of the Button, but with the new styles builder they can't do `.vtex-button span` because we don't allow to style direct elements, so we need to add classes to all elements of the store.

Before:
image

Now: 
image

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [9.105.0] - 2020-01-16
+
 ### Added
 - Handle to the label of the `Button`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+- Handle to the label of the `Button`.
+
 ## [9.104.11] - 2020-01-15
 
 ### Added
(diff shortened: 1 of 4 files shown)


emersonlaurentino:
I am not in favor of adding classes since our components are opinionated. If it were to use classes on the button, just pass a `className` to the component and time to create the style would be something like:

```css
/* label */
.buyButton span {
  display: none
}
```

But still I don't think we should be able to customize our components via css, one hour out of control.

**Maybe the cool way is to know what these agencies want to do with this label and then see if it makes sense to be a prop of the component itself.**

Klynger:
> I am not in favor of adding classes since our components are opinionated. If it were to use classes on the button, just pass a `className` to the component and time to create the style would be something like:
> 
> ```css
> /* label */
> .buyButton span {
>   display: none
> }
> ```
> 
> But still I don't think we should be able to customize our components via css, one hour out of control.
> 
> **Maybe the cool way is to know what these agencies want to do with this label and then see if it makes sense to be a prop of the component itself.**

Thats the problem, with the new styles builder we don't allow this kind of rule (styling html elements directly) so they can't do this anymore

emersonlaurentino:
but label has no classes precisely because it should not be customized. This would be nice to discuss at the Styleguide meeting, because once we use classes for the components they no longer have an interface pattern.

Klynger:
> but label has no classes precisely because it should not be customized. This would be nice to discuss at the Styleguide meeting, because once we use classes for the components they no longer have an interface pattern.

Yes, it shouldn't but thats a really old problem, the store shouldn't be using the styleguide at all because it has a totally different idea, but since we're using the styleguide button we need to support this kind of case

PR #1001, opened 2020-01-21
Allow the use of defaultMenuIsOpen in EXPERIMENTAL_Select

rerissondaniel:

As the title says

Check the use of the prop in the `Status` filter.

[workspace using new prop](https://addactivefilter--storecomponents.myvtex.com/admin/app/collections/1164/)
[workspace not using new prop](https://addactivefilter1--storecomponents.myvtex.com/admin/app/collections/1164/)

*Before*
chrome-capture

*After*
chrome-capture (1)

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [9.105.4] - 2020-01-22
+
 ### Added
 
 - `defaultMenuIsOpen` prop in `EXPERIMENTAL_Select`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
-### Fixed
+### Added
 
-- Now `defaultMenuIsOpen` can be used in `EXPERIMENTAL_Select`.
+- `defaultMenuIsOpen` prop in `EXPERIMENTAL_Select`.
 
 ## [9.105.3] - 2020-01-22
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+- Now `defaultMenuIsOpen` can be used in `EXPERIMENTAL_Select`.
+
 ## [9.105.3] - 2020-01-22
 
 ### Fixed
--- a/react/components/EXPERIMENTAL_Select/index.js
+++ b/react/components/EXPERIMENTAL_Select/index.js
@@ -260,7 +260,7 @@ Select.propTypes = {
   value: PropTypes.oneOfType([OptionShape, OptionsShape]),
   /** Max height (in _px_) of the selected values container */
   valuesMaxHeight: PropTypes.number,
-  /** If its options are initially shown*/
+  /** If its options are initially shown */
   defaultMenuIsOpen: PropTypes.bool,
 }
 
--- a/react/components/EXPERIMENTAL_Select/index.js
+++ b/react/components/EXPERIMENTAL_Select/index.js
@@ -60,9 +60,11 @@ class Select extends Component {
       value,
       valuesMaxHeight,
       clearable,
+      defaultMenuIsOpen
     } = this.props
 
     const reactSelectComponentProps = {
+      defaultMenuIsOpen,
       ref: forwardedRef,
       autoFocus,
       className: `pointer b--danger bw1 ${getFontClassNameFromSize(size)} ${
--- a/react/components/EXPERIMENTAL_Select/index.js
+++ b/react/components/EXPERIMENTAL_Select/index.js
@@ -197,6 +197,7 @@ Select.defaultProps = {
   placeholder: 'Select...',
   size: 'regular',
   clearable: true,
+  defaultMenuIsOpen: false,
 }
 
 const OptionShape = PropTypes.shape({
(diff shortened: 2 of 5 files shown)


julioleitao:
@rerissondaniel, can you add the gif with the previous and changed to be documented here?

rerissondaniel:
> @rerissondaniel, can you add the gif with the previous and changed to be documented here?

Done

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
```
kevinch: Shouldn't this a `Added` section?

kevinch: `defaultMenuIsOpen` option for `EXPERIMENTAL_Select`

react/components/EXPERIMENTAL_Select/index.js
```diff
@@ -258,6 +260,8 @@ Select.propTypes = {
   value: PropTypes.oneOfType([OptionShape, OptionsShape]),
   /** Max height (in _px_) of the selected values container */
   valuesMaxHeight: PropTypes.number,
+  /** If its options are initially shown*/
```
kevinch:
```suggestion
  /** If its options are initially shown */
```

davi-shama:
Hey, @rerissondaniel Shouldn't we be using the Fabriga font in this?

CHANGELOG.md
rerissondaniel: Yes, but it is a prop of the used library, so maybe it should be available... as not all props from the library are available, I'll use an added, makes more sense

rerissondaniel:
> Hey, @rerissondaniel Shouldn't we be using the Fabriga font in this?

Yes @davi-shama , but its not related to this change. Apparently in computers that don't have that font its not loaded by the browser (You can check that in the workspace without that change the font is the same)

davi-shama:
> > Hey, @rerissondaniel Shouldn't we be using the Fabriga font in this?
> 
> Yes @davi-shama , but its not related to this change. Apparently in computers that don't have that font its not loaded by the browser (You can check that in the workspace without that change the font is the same)

Great, thanks! Related to the component: do you have a next step about the Apply button? I see the improvement you are doing here, and it's moving foward. But I think that if we have only one select in the filter, as the example you showned, the filter itself should be the selector. 

Doing it super super poorly just to show:
image

rerissondaniel:
> > > Hey, @rerissondaniel Shouldn't we be using the Fabriga font in this?
> > 
> > 
> > Yes @davi-shama , but its not related to this change. Apparently in computers that don't have that font its not loaded by the browser (You can check that in the workspace without that change the font is the same)
> 
> Great, thanks! Related to the component: do you have a next step about the Apply button? I see the improvement you are doing here, and it's moving foward. But I think that if we have only one select in the filter, as the example you showned, the filter itself should be the selector.
> 
> Doing it super super poorly just to show:
> image

That's great! But should be done in the app using it (in this case admin collections)

davi-shama:
> > > > Hey, @rerissondaniel Shouldn't we be using the Fabriga font in this?
> > > 
> > > 
> > > Yes @davi-shama , but its not related to this change. Apparently in computers that don't have that font its not loaded by the browser (You can check that in the workspace without that change the font is the same)
> > 
> > 
> > Great, thanks! Related to the component: do you have a next step about the Apply button? I see the improvement you are doing here, and it's moving foward. But I think that if we have only one select in the filter, as the example you showned, the filter itself should be the selector.
> > Doing it super super poorly just to show:
> > image
> 
> That's great! But should be done in the app using it (in this case admin collections)

Is it possible to do it right now? I thought it wasn't. Do you have plans on removing the necessity of the Apply button? Saying all of that because I don't know if this PR is the best approach for minimizing user clicks. Clicking on a select and opening another one.

react/components/EXPERIMENTAL_Select/index.js
```diff
@@ -258,6 +260,8 @@ Select.propTypes = {
   value: PropTypes.oneOfType([OptionShape, OptionsShape]),
   /** Max height (in _px_) of the selected values container */
   valuesMaxHeight: PropTypes.number,
+  /** If its options are initially shown */
+  defaultMenuIsOpen: PropTypes.bool,
```
emersonlaurentino: shouldn't it be added to `defaultProps` with a value of `false`?

rerissondaniel: As its not required I think it doesn't matter, no?

emersonlaurentino: It's just a way of guaranteeing, because it will arrive undefined, but the code understands that it is a falsy.

rerissondaniel:
> > > > > Hey, @rerissondaniel Shouldn't we be using the Fabriga font in this?
> > > > 
> > > > 
> > > > Yes @davi-shama , but its not related to this change. Apparently in computers that don't have that font its not loaded by the browser (You can check that in the workspace without that change the font is the same)
> > > 
> > > 
> > > Great, thanks! Related to the component: do you have a next step about the Apply button? I see the improvement you are doing here, and it's moving foward. But I think that if we have only one select in the filter, as the example you showned, the filter itself should be the selector.
> > > Doing it super super poorly just to show:
> > > image
> > 
> > 
> > That's great! But should be done in the app using it (in this case admin collections)
> 
> Is it possible to do it right now? I thought it wasn't. Do you have plans on removing the necessity of the Apply button? Saying all of that because I don't know if this PR is the best approach for minimizing user clicks. Clicking on a select and opening another one.

Yes, but we would have to to it for the other filters (to keep it consistent). I Agree with but I should check with our designer. This PR only aims to allow the user of the select to set it to be initially open.

react/components/EXPERIMENTAL_Select/index.js
rerissondaniel: I'll add it anyway

PR #1010, opened 2020-01-23
Add size prop to checkbox

### LINKED ISSUE #799: Checkbox label doesn't allow font size small
**Describe the bug**
Checkbox label doesn't allow font size small (14pt). It should allow it because it makes sense in some scenarios, such as when regular/small inputs with font size small are already being used in a form.

**To Reproduce**
Open the styleguide documentation and see that this isn't possible in the current implementation.

**Expected behavior**
The component should allow small font size instead of body whenever needed.

**Screenshots**
Group 13

arthurepc:

Fix #799 

Run this project locally and go to the Checkbox page. Make sure that when the `size` prop is passed, the text size changes.

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 
-- size prop to checkbox.
+- Size prop to checkbox.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- size prop to checkbox.
+
 ### Changed
 
 - `Table`'s `headerRenderer` can be used in columns that are either fixed or sortable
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- size prop to checkbox.
```
bivillar:
```suggestion
- Size prop to checkbox.
```

react/components/Checkbox/index.js
```diff
@@ -119,6 +121,8 @@ Checkbox.propTypes = {
   value: PropTypes.string,
   /** Partial state */
   partial: PropTypes.bool,
+  /** Label size */
+  size: PropTypes.oneOf(['small']),
```
bivillar: Wouldn't it be better if this was a boolean prop named `small`?

kevinch: Not sure @bivillar since in the future we could add a "regular" or "large" for x reasons.

emersonlaurentino: I agree with what @kevinch said

PR #1012, opened 2020-01-24
Fix Errors in Table

emersonlaurentino:
There are some errors in the `Table` that are displayed on the `console`. The idea of ​​this PR is to make the console as clean as possible by correcting bugs.

- The `SwitchablePagination` component receives a prop called `enabled` which expects it to be of type `boolean`, however it was receiving an `object`.
- Props are required for the `Pagination` component but should not be required for the `SwitchablePagination` component since it renders only if `enabled` is` true`.

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
-- checks if the paginationClone object is not empty in `Table`.
-- remove propTypes required in `SwitchablePagination`.
+- Checks if the paginationClone object is not empty in `Table`.
+
+### Changed
+
+- PropTypes required in `SwitchablePagination`.
 
 ## [9.105.5] - 2020-01-23
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [9.106.1] - 2020-01-27
+
 ### Fixed
 
 - Checks if the paginationClone object is not empty in `Table`.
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+- checks if the paginationClone object is not empty in `Table`.
```
kevinch: Sounds to me that the sections should be `Changed` and `Removed`. What do you think?

react/components/Table/SwitchablePagination.js
```diff
@@ -15,8 +15,16 @@ SwitchablePagination.defaultProps = {
 }
 
 SwitchablePagination.propTypes = {
-  ...Pagination.propTypes,
```
kevinch: Where was it required? ```remove propTypes required in `SwitchablePagination`.```.

emersonlaurentino: the `currentItemFrom`, `currentItemTo`, `textOf`, `textShowRows`, `totalItems` props are required in pagination, but should not in `SwitchablePagination`.

emersonlaurentino: and I can't remove the required ones inside the required component, because there it makes sense to be. So the spread on all propTypes of the pagination does not make sense for SwitchablePagination.

emersonlaurentino: that's why I added here only the props that are actually used.

kevinch: Ok.

CHANGELOG.md
emersonlaurentino: maybe in `remove propTypes required in SwitchablePagination.` better use the `Changed`. But for `checks if the paginationClone object is not empty in Table.` makes sense to use `Fixed`

PR #1021, opened 2020-01-30
Feature/radiogroup label

steckhelena:

The purpose of this PR is adding the possibility of adding a label to a RadioGroup

It would be needed for labels to be put manually if you wanted to label a RadioGroup before and also now with the fieldset, it is better for screen readers.

image

#### Types of changes

### DIFF
--- a/react/components/RadioGroup/index.js
+++ b/react/components/RadioGroup/index.js
@@ -25,10 +25,7 @@ class RadioGroup extends React.Component {
 
     return (
       <div>
-        <fieldset
-          className="vtex-input w-100"
-          data-testid={testId}
-          style={{ border: 'none', padding: 0 }}>
+        <fieldset className="vtex-input w-100 bn pa0" data-testid={testId}>
           <legend>
             {label && (
               <span
--- a/react/components/RadioGroup/index.js
+++ b/react/components/RadioGroup/index.js
@@ -20,19 +20,9 @@ class RadioGroup extends React.Component {
       testId,
       size,
     } = this.props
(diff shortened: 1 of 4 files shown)


react/components/RadioGroup/index.js
```diff
@@ -10,48 +10,77 @@ class RadioGroup extends React.Component {
...
+      case 'small':
+        labelClasses += 't-small '
+        break
+      case 'large':
+        labelClasses += 't-body '
+        break
+      default:
+        labelClasses += 't-small '
+        break
+    }
```
matheusps: You can use `classnames` for that. Example: 

```js
import classnames from 'classnames'

const large = size === 'large'

const labelClasses = classnames('vtex-input__label db mb3 w-100 c-on-base', {
  't-body': large,
  't-small': !large
})
```

steckhelena: Thanks for the input! I implemented your sugestions!

react/components/RadioGroup/index.js
```diff
@@ -10,48 +10,75 @@ class RadioGroup extends React.Component {
...
-                />
-              </div>
-            </label>
-          )
-        })}
+        <fieldset
+          className="vtex-input w-100"
+          data-testid={testId}
+          style={{ border: 'none', padding: 0 }}>
+          <legend>
```
kaisermann: Should the `<legend>` element exist if there's no `label`?

steckhelena: Fixed it!

PR #1026, opened 2020-02-03
Chore/components index

thiagomurakami:
- Add the file `react/components/index.js` so users installing from NPM can import components like they import components in IO; i.e.:
``` typescript
import { Button } from '@vtex/styleguide'
// instead of
import Button from '@vtex/styleguide/lib/Button'
```

Also, fixing imports in the `ColorPicker` component.

1. `ColorPicker` is broken when using from NPM: [CodeSandbox](https://codesandbox.io/s/goofy-perlman-5md7f)

I'm currently installing `storybook` in a new VTEX IO app and I need to locally mock `vtex.styleguide` (by using [resolve.alias](https://webpack.js.org/configuration/resolve/#resolvealias)) and the best way to achieve this is by having `@vtex/styleguide` have the same API as `vtex.styleguide`.

`yarn babel:lib`
`yarn link @vtex/styleguide`

#### Types of changes


package.json
```diff
@@ -98,6 +99,7 @@
     "react-sticky": "^6.0.3",
     "react-virtualized": "^9.19.1",
     "recharts": "^2.0.0-beta.1",
+    "styled-components": "^5.0.0",
```
emersonlaurentino: why?

thiagomurakami: `styled-components` imports in `EXPERIMENTAL_Table` were breaking without this dependency on root, e.g., [Cell.tsx:2](https://github.com/vtex/styleguide/blob/master/react/components/EXPERIMENTAL_Table/DataTable/Cell.tsx#L2)

emersonlaurentino: wow, I didn't know we were using styled-components. This should not have happened.

PR #1029, opened 2020-02-04
Add modalClassName to Modal component

rerissondaniel:

Allow the user to provide classNames to `Modal` component. The modal width is too large for some cases and not flexible. This allows its users to customize that and other styles through classnames.

[Running workspace](https://modalcreatecol--cosmetics1.myvtex.com/admin/app/collections/create)

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- `modalClassName` prop to modal.
+
 ## [9.108.8] - 2020-02-04
 
 ### Changed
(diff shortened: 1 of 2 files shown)


emersonlaurentino:
we already had this discussion so that the components would receive a prop `className` however it was denied for the reason that if it allowed this, the components would no longer be strongly opinionated.

Why do you need a new className on Modal? Change the size of it? Does this exist in another context or is it specific to you?

matheusps:
Would it not possible to have size variations that are strongly opinionated? 🤔(Ex: `small`, `medium`, `large`, `full`, etc...)

emersonlaurentino:
@matheusps if the idea is to change the size of the modal, the ideal would be to have a prop just for that. My question is that today Idk places that have different sizes for modal, this would have to be aligned between designers.

@rerissondaniel Once the designers say that it makes sense to have different sizes for modals, create a prop `size` with the possible different sizes.

kaisermann:
accidentally approved, sorry 🤓

PR #1030, opened 2020-02-05
Add test id for MoreOptions button

mariana-mendes:
as the title says.

Avoid use css selectors in test implementations.

[Workspace](https://mariana--cosmetics1.myvtex.com/admin/collections/1006/).
#### Types of changes

### DIFF
--- a/react/components/FilterBar/index.js
+++ b/react/components/FilterBar/index.js
@@ -198,7 +198,7 @@ FilterBar.defaultProps = {
   submitFilterLabel: 'Apply',
   newFilterLabel: 'New Filter',
   statements: [],
-  testIds: {}
+  testIds: {},
 }
 
 export const filterBarPropTypes = {
@@ -220,10 +220,10 @@ export const filterBarPropTypes = {
   submitFilterLabel: PropTypes.string,
   /** New Filter title label for inside the 'More options' menu */
   newFilterLabel: PropTypes.string,
-  testIds:  PropTypes.shape({
+  testIds: PropTypes.shape({
     moreOptionsButton: PropTypes.string,
-    submitFiltersButton: PropTypes.string
-  })
+    submitFiltersButton: PropTypes.string,
+  }),
 }
 
 FilterBar.propTypes = filterBarPropTypes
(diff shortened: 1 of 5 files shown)


react/components/FilterBar/index.js
```diff
@@ -217,6 +219,7 @@ export const filterBarPropTypes = {
   submitFilterLabel: PropTypes.string,
   /** New Filter title label for inside the 'More options' menu */
   newFilterLabel: PropTypes.string,
+  testIdMoreOptions: PropTypes.string,
```
matheusps: Why do we need to pass it by props every time? 🤔. I think the internal components should have static ids.

mariana-mendes: Because we can use the component more than once in a app, then, they should have a unique key, the id.

mariana-mendes:
> Can't we use the other queries?
> 
> Test Id should be the last resort:
> https://testing-library.com/docs/guide-which-query

The intention is use native queries from cypress. :thinking:

emersonlaurentino:
we already had a discussion about this in #848 and we did some PRs in #856 

in short: it would be to use the `testId` prop and the attribute in the element to be `data-testid` because it is an attribute that can be read by Cypress and the Testing Library.

react/components/FilterBar/index.js
matheusps: Got it, makes sense.

react/components/FilterBar/index.js
```diff
@@ -217,6 +220,10 @@ export const filterBarPropTypes = {
   submitFilterLabel: PropTypes.string,
   /** New Filter title label for inside the 'More options' menu */
   newFilterLabel: PropTypes.string,
+  testIds: PropTypes.shape({
+    moreOptionsButton: PropTypes.string,
+    submitFiltersButton: PropTypes.string,
+  }),
```
rerissondaniel: These should be inside a `moreOptionsFilterTag` object, no? This way if we wanted to add ids to clearAllFilterButton we could only add a new prop to testIds

PR #1033, opened 2020-02-10
Fix highlight color of a selected row

mariana-mendes:
As the title says.

The color of a secondary button clicked should be `#DBE9FD (light blue)`

[Running workspace](https://mariana--cosmetics1.myvtex.com/admin/collections/1019/)

* Before: 
image

* After: 
image

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
+
+- Color of active row on `EXPERIMENTAL_Table`.
+
 ## [9.112.2] - 2020-02-27
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


mariana-mendes:
> @mariana-mendes `bg-action-secondary` is being used correctly. The color of the button secondary when clicked cannot be used, because it is not a token in tachyons it is a pseudo-class. And if we use constant colors, once we want to change our palette in tachyons ... it would break this component.
> 
> ```css
> .active-bg-action-secondary:active { background-color: #dbe9fd; }
> ```

`.active-bg-action-secondary:active { background-color: #dbe9fd; }` doesn't works for me, because `active` is a temporary action... So, if I use this class, I will lose the right color on the refresh page for example.
Actually, a solution like that is begin used  in `FilterTag` component :sweat_smile: 

Should I create a new class on `vtex-tachyons`? What do you think?

Issue #1043, opened 2020-02-17
Put recharts to PeerDependencies


emersonlaurentino:
Currently `recharts` is responsible for 21.2% of the total styleguide size.

Since our charts components are still experimental and are very specific apps that use them, it makes sense that `recharts` will be a dependency on these apps and a PeerDependence of the styleguide.

<img width="1372" alt="Screen Shot 2020-02-17 at 11 57 17" src="https://user-images.githubusercontent.com/10627086/74664527-a7216680-517c-11ea-9b74-0d7e695c11d5.png">

ref: [Bundlephobia](https://bundlephobia.com/result?p=@vtex/styleguide@9.111.0)

PR #1046, opened 2020-02-17
Update react dependency and add as peer

lucasecdb:
Updates the `react` and `react-dom` dependency (for development) and adds them as peerDependencies.

When you import styleguide in a normal react app, you may get the following error:

```
Error: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
See https://fb.me/react-invalid-hook-call for tips about how to debug and fix this problem.
```

I believe this is due to react not being present in the peerDependencies, so when webpack compiles this to publish on NPM it insert it's own copy of React in the package, making any usage of hooks throw when using any component. This does not happen on IO because we build the app in a different way (not using the webpack config in this repo, but one in builder-hub).

Create a react app with `yarn create react-app styleguide-test`. Add styleguide as a dependency and try to use any of it's components.

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+- Packages `react@^16.8.0` and `react-dom@^16.8.0` to `peerDependencies`.
+
 ## [9.111.3] - 2020-02-17
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [9.111.4] - 2020-02-17
+
 ### Added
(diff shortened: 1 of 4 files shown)


emersonlaurentino:
@lucasecdb great job.

it would be cool to resolve this issue https://github.com/vtex/styleguide/issues/1043 in this PR as well.

lucasecdb:
@emersonlaurentino my bad! I forgot to update the changelog

lucasecdb:
don't you think it's better to do this update in a separate PR then? This is blocking some tasks and I don't wan't to wait.

emersonlaurentino:
@lucasecdb do you need these modifications to react and react-dom urgently?

lucasecdb:
> @lucasecdb do you need these modifications to react and react-dom urgently?

Yes, because this is blocking for some tasks that are on hold since last week, and we need to update them.

emersonlaurentino:
Vish, then separate into another PR.

PR #1047, opened 2020-02-17
Added sticky header to EXPERIMENTAL_Table

matheusps:

As the title says

The Table headers can be fixed for a better viz when the items list is large.
This PR also removes the possibility of passing different elements then `td`, `th`, `tr` or `table`. This is to grant accessibility.

Clone the repo and `yarn && yarn start`

Screen Recording 2020-02-17 at 17 07 05 2020-02-17 17_07_51

#### Types of changes

### DIFF
--- a/react/components/EXPERIMENTAL_Table/DataTable/Cell.tsx
+++ b/react/components/EXPERIMENTAL_Table/DataTable/Cell.tsx
@@ -35,19 +35,31 @@ const Cell: FC<CellProps> & CellComposites = ({
   className: classNameProp = '',
   active = false,
   link = false,
+  sticky = false,
+  header,
 }) => {
   const { hover, ...events } = useHover()
   const className = classNames(
-    'v-mid ph3 pv0 tl bb b--muted-4',
+    'v-mid ph3 pv0 tl bb b--muted-4 bg-base',
     classNameProp,
     {
       pointer: onClick,
       'hover-c-link hover-bg-muted-5': link,
       'c-on-base': active,
+      'top-0 z3': sticky && header,
+      z1: !sticky,
     }
   )
+
   return (
-    <Tag {...events} onClick={onClick} style={{ width }} className={className}>
+    <Tag
+      {...events}
+      onClick={onClick}
+      style={{
+        position: sticky ? 'sticky' : 'static',
+        width,
+      }}
+      className={className}>
       <HoverProvider value={hover}>{children}</HoverProvider>
     </Tag>
   )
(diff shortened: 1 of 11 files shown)


react/components/EXPERIMENTAL_Table/DataTable/Cell.tsx
```diff
@@ -35,19 +35,31 @@ const Cell: FC<CellProps> & CellComposites = ({
   className: classNameProp = '',
   active = false,
   link = false,
+  sticky = false,
+  header,
```
emersonlaurentino: why is this prop necessary?

react/components/EXPERIMENTAL_Table/DataTable/Cell.tsx
```diff
@@ -35,19 +35,31 @@ const Cell: FC<CellProps> & CellComposites = ({
...
     }
   )
+
   return (
-    <Tag {...events} onClick={onClick} style={{ width }} className={className}>
+    <Tag
+      {...events}
+      onClick={onClick}
+      style={{
+        position: sticky ? 'sticky' : 'static',
```
Klynger: I don't know if we should be doing this using css because of the [support of browsers](https://caniuse.com/#search=sticky). In the store we had to do something similar, so we create this behavior with js. https://github.com/vtex-apps/sticky-layout cc @kaisermann

kaisermann: Yeah, support for actual tables is not great, but it seems you're not using `<tr/td/thead>`, right?

matheusps: I only use the `sticky` position on the `<th/>`s.  That's the only table element that works well with this feature on stable browsers. Tested on chrome, safari, firefox, and edge (for mac).

react/components/EXPERIMENTAL_Table/DataTable/Cell.tsx
matheusps: updated.

emersonlaurentino: @matheusps why `header` is necessary?

matheusps: Ops. I edited my first answer, for some reason just 'updated' stayed 😅(sorry). The header defines if is a `th` or `td`. And we may not use `sticky` on a `td`.

PR #1048, opened 2020-02-17
Fix/recharts

emersonlaurentino:
fix: https://github.com/vtex/styleguide/issues/1043

fix: https://github.com/vtex/styleguide/issues/1043

check size in bundlephobia

[@vtex/styleguide@9.111.0](https://bundlephobia.com/result?p=@vtex/styleguide@9.111.0)

#### Types of changes

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -2,7 +2,7 @@
   "vendor": "vtex",
   "name": "styleguide",
   "title": "VTEX Styleguide",
-  "version": "9.111.5-beta",
+  "version": "9.111.5",
   "description": "The VTEX Styleguide components for the Render framework",
   "builders": {
     "react": "3.x"
--- a/manifest.json
+++ b/manifest.json
@@ -2,7 +2,7 @@
   "vendor": "vtex",
   "name": "styleguide",
   "title": "VTEX Styleguide",
-  "version": "9.111.4",
+  "version": "9.111.5-beta",
   "description": "The VTEX Styleguide components for the Render framework",
   "builders": {
     "react": "3.x"
--- a/react/package.json
+++ b/react/package.json
@@ -4,7 +4,6 @@
     "lodash": "^4.17.11",
     "react-color": "^2.17.0",
     "react-datepicker": "^2.0.0",
-    "react-dom": "^16.9.0",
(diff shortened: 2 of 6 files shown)


manifest.json
```diff
@@ -2,7 +2,7 @@
   "vendor": "vtex",
   "name": "styleguide",
   "title": "VTEX Styleguide",
-  "version": "9.111.4",
+  "version": "9.111.5-beta",
```
arthurepc: Do you have any specific reason for this to be a beta tag?

emersonlaurentino: test size in bundlephobia

https://bundlephobia.com/result?p=@vtex/styleguide@9.111.5-beta

emersonlaurentino: but `releasy` has a problem, he publish beta as latest!

emersonlaurentino: Screen Shot 2020-02-17 at 19 28 52

emersonlaurentino:
We have a problem, if we put `recharts` as peerDependencies solves our problem with the size of the lib, however our documentation will break because we will not be able to find the dependency, consequently we would need to add `recharts` to devDependencies, which would make the lib big again.

react/package.json
```diff
@@ -26,6 +24,12 @@
     "@types/node": "^12.6.2",
     "@types/react": "^16.8.23",
     "@types/react-dom": "^16.9.0",
-    "@types/recharts": "^1.8.0"
+    "@types/recharts": "^1.8.0",
+    "recharts": "^2.0.0-beta.1"
```
arthurepc: Also, do you have some specific reason to use this beta version of recharts?

arthurepc: And should this be in both `devDependencies` and `peerDependencies`?

emersonlaurentino: > Also, do you have some specific reason to use this beta version of recharts?

The beta version was already being used, the Fanny used it, so I didn't want to change the version, just put it as peerDependencies.

> And should this be in both devDependencies and peerDependencies?

I explained in my last comment https://github.com/vtex/styleguide/pull/1048#issuecomment-587179160

emersonlaurentino: the ideal scenario would be for our documentation to be separate from our styleguide.

PR #1049, opened 2020-02-18
Move postinstall script to prepare script

lucasecdb:
Moves the `postinstall` script to the [`prepare`](https://docs.npmjs.com/misc/scripts#prepublish-and-prepare) script, since it's more appropriate.

This is to avoid adding the dev dependencies of this package in production environments, which also breaks React Hooks because styleguide will be using it's own version of React, which will be different than the one used in the application importing it.

Create a sample react project with

```
yarn create react-app styleguide-test
cd styleguide-test
```

Then, add this version of styleguide and verify if any component that uses hooks is rendering correctly. **It is important to *not* run `yarn` on the styleguide directory, as it will always bring the node_modules folder when you locally `yarn add` it.**

```
yarn add ../path/to/styleguide
yarn start
```

```jsx
// in App.js
import React from 'react'
import ButtonPlain from '@vtex/styleguide/lib/ButtonPlain'

export default function App() {
  return <ButtonPlain>Click me</ButtonPlain>
}
```

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
+
+- Added `--prod` flag to `postinstall` script.
+
 ## [9.111.5] - 2020-02-17
 
 ## [9.111.5-beta] - 2020-02-17
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 
-- Added `--prod` flag to `postinstall` script.
+- Moved the  `postinstall` script to `prepare`, to avoid adding devDependencies on production.
 
 ## [9.111.5] - 2020-02-17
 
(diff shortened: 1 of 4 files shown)


emersonlaurentino:
@lucasecdb in fact `prepare` is more appropriate for this case, since we just want to install what is in the package in the react folder, however, I didn’t understand how this solves devDependencies.

lucasecdb:
@emersonlaurentino The issue was that with the `postinstall` script, the package manager (either yarn or npm) was executing this when the styleguide package was *installed* (i.e. when you `yarn add`ed `@vtex/styleguide`), which then triggered the `yarn install` inside the `node_modules/@vtex/styleguide`. This installed all of styleguide's dependencies, including their dev dependencies, which caused all the issues seen with duplicate React copies, etc.

Moving it from `postinstall` to `prepare`, the script will only execute now on the dev which is *working on* styleguide, instead of the dev who is *using* styleguide. The postinstall script is used on packages like `node-sass` to compile the sass binaries to the CPU architecture of who is using it, which makes sense for them, but isn't really the case for us.

hope this clarifies more the intent of this PR :-)

emersonlaurentino:
I understand, you could have solved this by removing react from react/package.json since it is already present in the package that is at the root of the project. This would solve the problem with different versions (duplicates).

I think you can use `prepare` and remove the react from `react/package.json`

lucasecdb:
no, the issue was that the react in the `devDependencies` of the root `package.json` was being installed on the styleguide directory inside `node_modules` when used with NPM, not VTEX IO. I don't think removing `react` from the `react/package.json` file would make much difference, because that file is compiled away when pushing the assets to the NPM registry.

PR #1040, opened 2020-02-14
Adds LocationIcon

mredigonda:
#### What is the purpose of this pull request?

Adds a LocationIcon needed for `PlaceComponents`.

#### What problem is this solving?

The `DeviceCoordinates` component (from `vtex-apps/place-components`) requires these icons, and instead of defining them there, we decided to share them in the styleguide since they might be needed by other projects too.

#### How should this be manually tested?

In your VTEX IO app, try to import the newly added icon with `import { IconLocation } from 'vtex.styleguide'`, and then use it like: `<IconLocation />` or `<IconLocation granted />`. Test that you can modify its size and color like with most other icons.

[Workspace](https://styleguidelocation--checkoutio.myvtex.com/places). The `DeviceCoordinates` component should be displaying the icon.

#### Screenshots or example usage

image

#### Types of changes

- [ ] Bug fix (a non-breaking change which fixes an issue)
- [x] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Requires change to documentation, which has been updated accordingly.

PR #1057, opened 2020-02-28
Release v9.112.5

emersonlaurentino:
What is the purpose of this pull request?
Add missing tag/release forgotten on previous PR (#1040)


arthurepc:
Also, I think we should have double the attention when releasing something in a friday, specially while demo friday is happening.

PR #1058, opened 2020-02-28
Adds `buttonProps` properties to `Input` and `InputButton`

mredigonda:

Currently, the `InputButton` styleguide component does not allow you to pass props to the button it renders. The reason for this is that the component was meant to be inside of a `form`, however, the `Location Field (Input)` Place Component needs to render a `InputButton` without the context of a form, since that component in itself can be embedded into another form, which would create a form inside of a form.
The solution is to provide a way to add props to this button, so that one can pass actions like `onClick` to it, in such a way that you don't need a wrapper form to use the component.

The component is being used under the "LocationInput" title.

[Workspace](https://locinput--checkoutio.myvtex.com/places)

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [9.112.11] - 2020-03-06
+
 ### Added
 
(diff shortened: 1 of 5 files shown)


emersonlaurentino:
@mredigonda dont forget to add these modifications in changelog.

Issue #1061, opened 2020-03-03
Migrate Button component to typescript


igorbrasileiro:
## Purpose

Migrate Button component to typescript

emersonlaurentino:
remember to pull request to the `dev` branch

kevinch:
Sounds like this issue can be closed since we have https://github.com/vtex/styleguide/issues/1084

PR #1066, opened 2020-03-05
Migrate Button to Typescript and convert to function component

igorbrasileiro:
As the title says. Related issue: #1061 

#### Types of changes


react/modules/withForwardedRef.tsx
```diff
@@ -1,21 +1,23 @@
...
+  ComponentType,
+} from 'react'
 import PropTypes from 'prop-types'
 
 // For more info see: https://stackoverflow.com/a/51127130/10725088
-const Element =
-  typeof window === 'undefined' || typeof window.Element === 'undefined'
-    ? function() {}
-    : window.Element
-
```
emersonlaurentino: why did you delete that?

react/components/Button/index.tsx
```diff
@@ -1,11 +1,124 @@
...
+  icon: false,
+  type: 'button',
+  isLoading: false,
+  isGrouped: false,
+  isFirstOfGroup: false,
+  isLastOfGroup: false,
+  isActiveOfGroup: false,
+}
+
+class Button extends Component<InferProps<typeof propTypes>> {
```
matheusps: This could be an `FC` since you are upgrading.

igorbrasileiro: Ok @matheusps !

react/modules/withForwardedRef.tsx
klzns: That change will break SSR

igorbrasileiro: Yeah, I put it back!

Issue #938, opened 2019-11-12
Error messages in fields need an icon for color-blind users


davicosta99:
**Describe the bug**
Outlining the field red and adding a message in red is not enough. Color-blind users need an icon to go along with the message.

**Screenshots**
![1_F7OGfYAElevZDXpyb-QzIw](https://user-images.githubusercontent.com/8797476/68691708-f2f7ba80-0552-11ea-9663-90192dc7f89a.png)

**Additional context**
My first read after discovering this issue is that help messages don't need the icon. They are lighter, and don't need to be as discoverable as an error message.

jonathasbsouza:
I've made an example[ here](https://www.figma.com/file/rnJzqiDzLZ86yicLCrSYIB/02-Forms?node-id=106%3A146) using the Warning icon at 16x16px using the same color as the message text.

**Protanopia**
<img width="314" alt="Screen Shot 2020-03-05 at 16 34 04" src="https://user-images.githubusercontent.com/7189694/76024314-9fe2b280-5f09-11ea-8327-764715c3a952.png">

**Deuteranopia**
<img width="308" alt="Screen Shot 2020-03-05 at 16 34 10" src="https://user-images.githubusercontent.com/7189694/76024320-a2dda300-5f09-11ea-81ad-11f147f02249.png">

**Tritanopia**
<img width="292" alt="Screen Shot 2020-03-05 at 16 34 17" src="https://user-images.githubusercontent.com/7189694/76024325-a53ffd00-5f09-11ea-8043-85776f8cd4db.png">

emersonlaurentino:
closing this issue, because it was resolved in #1077

PR #1069, opened 2020-03-06
Add warning icon when input has errorMessage

igorbrasileiro:
As the title says

issue: #938 

Use an input and pass an errorMessage prop

**before**
image

**after**
image

#### Types of changes

### DIFF
--- a/react/components/Input/index.js
+++ b/react/components/Input/index.js
@@ -263,11 +263,12 @@ class Input extends Component {
           )}
         </div>
         {errorMessage && (
-          <div className="c-danger flex t-small mt3 lh-title">
-            <span className="mr2 flex items-center">
+          <div className="c-danger flex items-center t-small mt3 lh-title">
+            {/* use flex to align svg vertically */}
+            <span className="mr2 flex">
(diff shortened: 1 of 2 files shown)


react/components/Input/index.js
```diff
@@ -262,7 +263,12 @@ class Input extends Component {
           )}
         </div>
         {errorMessage && (
-          <div className="c-danger t-small mt3 lh-title">{errorMessage}</div>
+          <div className="c-danger flex t-small mt3 lh-title">
+            <span className="mr2 flex items-center">
```
kaisermann: This doesn't need to be a `flex`, you can add the `item-scenter` to the parent div.

igorbrasileiro: You said to use `flex items-center` on div and remove `flex items-center` from span, right? I tried to use this way but the result wasn't the same.

igorbrasileiro: SVG element is `display: inline`,  so line-height and font-size affect it. If I use `display: flex` Svg display will be updated to `block` and won't be affected by line-height and font-size.

PR #1070, opened 2020-03-09
#trivial Feature tests

klzns:

Add snapshot tests to Button so we can keep track of changes in style and avoid breaking stores' customizations.

Motivation of this PR is to guarantee that #1068 doesn't break our Button.

> **There's no changelog change because it doesn't change our API**

Run tests 😅 

#### Types of changes

- [x] Changes in tests

### DIFF
--- a/react/components/Alert/index.test.js
+++ b/react/components/Alert/index.test.js
@@ -3,6 +3,8 @@ import { render, fireEvent } from 'react-testing-library'
 
 import Alert from './index'
 
+// Skipping those tests because getByText is getting the div inside the button and not the button element itself
+// I've tried to fix it, but I couldn't manage to. So I let me fellows future developers in this endeavour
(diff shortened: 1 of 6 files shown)


react/components/Alert/index.test.js
```diff
@@ -3,7 +3,7 @@ import { render, fireEvent } from 'react-testing-library'
 
 import Alert from './index'
 
-describe('Alert', () => {
+describe.skip('Alert', () => {
```
iaronaraujo: Why the skip?

klzns: image

and

image

iaronaraujo: There should be a comment above the skip then

iaronaraujo: Add a comment above the skip

PR #1073, opened 2020-03-10
Make new release since previous wasn't published

rerissondaniel:

As the title says

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+- Bump version to generate new release.
+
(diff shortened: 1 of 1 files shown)


emersonlaurentino:
Why? once the version exists. Just rotate the controls:
- `npm publish`
- `vtex publish`

the last two versions are in npm, can you check if they are in vtexio?

emersonlaurentino:
@rerissondaniel I checked here and the last two versions were also deployed to vtexio, it may just be necessary to do `vtex deploy`

rerissondaniel:
> @rerissondaniel I checked here and the last two versions were also deployed to vtexio, it may just be necessary to do `vtex deploy`

I've just ran vtex publish in head of master, so this is not necessary anymore, the .15 really wasn't published in IO.
In here it says it is in version 9.112.14
https://www.npmjs.com/package/@vtex/styleguide

PR #1087, opened 2020-03-13
Fix Dropdown usage inside an Input prefix or suffix

lucasecdb:
Adds a better support for the `Dropdown` component inside an `Input` prefix or suffix.

[Running workspace](https://lucas--checkoutio.myvtex.com/checkout/?cultureInfo=pt-BR#/profile)

With this, implementing the following design is more straightforward and requires no "hack" in userland.

image

#### Types of changes

### DIFF
--- a/react/components/Dropdown/Dropdown.css
+++ b/react/components/Dropdown/Dropdown.css
@@ -1,0 +1,28 @@
+:global(.vtex-input__prefix) .dropdown,
+:global(.vtex-input__suffix) .dropdown {
+  height: 100%;
+}
+
+:global(.vtex-input__prefix) .dropdown {
+  margin-left: -1rem;
+}
+
+:global(.vtex-input__suffix) .dropdown {
+  margin-right: -1rem;
+}
+
+:global(.vtex-input__prefix) .container,
+:global(.vtex-input__suffix) .container {
+  height: 100%;
+  border-top: none;
+  border-bottom: none;
+  border-radius: 0;
+}
+
+:global(.vtex-input__prefix) .container {
+  border-left: none;
+}
+
+:global(.vtex-input__suffix) .container {
+  border-right: none;
+}
--- a/react/components/Dropdown/index.js
+++ b/react/components/Dropdown/index.js
@@ -1,3 +1,4 @@
+import classNames from 'classnames'
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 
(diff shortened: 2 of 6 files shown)


lucasecdb:
If you get an error while accessing the page I linked to, please go [here](https://lucas--checkoutio.myvtex.com/checkout/identification?cultureInfo=pt-BR) first and input any email. This page is still an WIP and is not working in all scenarious yet.

react/components/Dropdown/index.js
```diff
@@ -110,7 +111,9 @@ class Dropdown extends Component {
     let iconSize
 
     let classes = 'bg-transparent bn w-100 h-100 '
-    let containerClasses = `${isInline ? '' : 'bw1'} br2 relative `
+    let containerClasses = `${styles.container} ${
+      isInline ? '' : 'bw1'
+    } br2 relative `
     let selectClasses = 'o-0 absolute top-0 left-0 h-100 w-100 bottom-00 '
 
     let labelClasses = 'vtex-dropdown__label db mb3 w-100 c-on-base '
```
emersonlaurentino: can u use `classNames`?

react/components/Dropdown/Dropdown.css
```diff
@@ -0,0 +1,28 @@
...
+  border-radius: 0;
+}
+
+:global(.vtex-input__prefix) .container {
+  border-left: none;
+}
+
+:global(.vtex-input__suffix) .container {
+  border-right: none;
+}
```
emersonlaurentino: why are all classes as global?

react/components/Dropdown/index.js
lucasecdb: didn't know I could! absolutely

react/components/Dropdown/Dropdown.css
lucasecdb: because this file is transpiled to css modules, and if I use them normally here (without the `:global`) I wouldn't be able to style it correctly

react/components/Dropdown/index.js
emersonlaurentino: the idea is that if we are able to use classes, we will use `classNames`.

lucasecdb: btw, by `classNames` you mean [this package](https://www.npmjs.com/package/classnames), right?

emersonlaurentino: Yes!

lucasecdb: done!

lucasecdb:
@emersonlaurentino the input's width is now breaking in some conditions. I'll try to revisit my changes before merging to avoid breaking existing layouts

lucasecdb:
fixed it! Could you look again @emersonlaurentino? 😅

PR #1088, opened 2020-03-13
Fix EXPERIMENTAL_useTableMeasures SSR errors

thayannevls:
Fixing this error:
Screenshot from 2020-03-13 11-24-12
 

[Fixed workspace](https://thay2--cosmetics1.myvtex.com/wedemo)

[Broken workspace](https://thay--cosmetics1.myvtex.com/wedemo)

#### Types of changes

### DIFF
--- a/react/components/EXPERIMENTAL_Table/hooks/useTableMeasures.tsx
+++ b/react/components/EXPERIMENTAL_Table/hooks/useTableMeasures.tsx
@@ -39,7 +39,9 @@ export function calculateTableHeight(
 }
 
 export function getScrollbarWidth(): number {
-  const isSSR = typeof document === 'undefined'
+  const isSSR =
+    typeof document === 'undefined' || typeof document === 'undefined'
+
   if (isSSR) {
     return DEFAULT_SCROLLBAR_WIDTH
   }
--- a/react/components/EXPERIMENTAL_Table/hooks/useTableMeasures.tsx
+++ b/react/components/EXPERIMENTAL_Table/hooks/useTableMeasures.tsx
@@ -39,8 +39,7 @@ export function calculateTableHeight(
 }
 
 export function getScrollbarWidth(): number {
-  const isSSR =
-    typeof document === 'undefined' || typeof document === 'undefined'
+  const isSSR = typeof document === 'undefined' || typeof window === 'undefined'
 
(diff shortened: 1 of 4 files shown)


react/components/EXPERIMENTAL_Table/hooks/useTableMeasures.tsx
```diff
@@ -41,8 +41,10 @@ export function calculateTableHeight(
 }
 
 export function getScrollbarWidth(): number {
-  if (!window || !document || !document.documentElement)
+  const isSSR = typeof document === 'undefined'
```
emersonlaurentino: Shouldn't it be like that?
```suggestion
  const isSSR = typeof document === undefined
```

matheusps: nope, it compares as a string [typeof ref](https://developer.mozilla.org/pt-BR/docs/Web/JavaScript/Reference/Operators/typeof)

thayannevls: The `typeof` operator always return a string with the type :smile:

lucasmedeiros: Shouldn't we check if `window` is also not set like that, or it isn't necessary?

```suggestion
  const isSSR = typeof document === 'undefined' || !window
```

thayannevls: Done, thanks :smile:

react/components/EXPERIMENTAL_Table/hooks/useTableMeasures.tsx
```diff
@@ -39,8 +39,12 @@ export function calculateTableHeight(
 }
 
 export function getScrollbarWidth(): number {
-  if (!window || !document || !document.documentElement)
+  const isSSR =
+    typeof document === 'undefined' || typeof document === 'undefined'
```
vitoriaHeliane: Why are you using the same terms in the logic expression?

thayannevls: True :woman_facepalming: I'll launch again fixing this

Issue #1082, opened 2020-03-12
Fix all eslint problems in the project


emersonlaurentino:
There are currently a lot of errors and warnings because of the eslint update in #1080.

The purpose of this issue is to fix all the eslint errors that currently exist so that issue #1081 can be resolved.

Files:
- [x] `/codemod/__tests__/button-v4.test.js`
- [x] `/codemod/button-v4.js` #1216 
- [x] `/config/env.js` #1092
- [x] `/config/jest/fileTransform.js` #1092
- [x] `/config/polyfills.js` #1092
- [x] `/config/test.js` #1092
- [x] `/config/webpack.config.js` #1092
- [x] `/react/EXPERIMENTAL_Conditions.js` #1214 
- [x] `/react/components/Alert/index.js` #1091
- [x] `/react/components/Alert/index.test.js` #1091
- [x] `/react/components/AutocompleteInput/SearchInput/index.tsx` #1231 
- [x] `/react/components/AutocompleteInput/index.tsx` #1231 
- [x] `/react/components/Button/index.test.tsx`#1118
- [x] `/react/components/Button/index.tsx`#1118
- [x] `/react/components/ButtonPlain/index.tsx` #1230 
- [x] `/react/components/ColorPicker/ColorHistory.js` #1207 
- [x] `/react/components/ColorPicker/HexInput.js` #1207 
- [x] `/react/components/ColorPicker/index.js` #1207 
- [x] `/react/components/Conditions/StrategySelector.tsx` #1167
- [x] `/react/components/Conditions/index.tsx` #1167
- [x] `/react/components/DatePicker/index.js` #1210 
- [x] `/react/components/Dropdown/index.js` #1210 
- [x] `/react/components/Dropzone/index.js` #1104
- [x] `/react/components/EXPERIMENTAL_Charts/BarChart/index.tsx` #1232 
- [x] `/react/components/EXPERIMENTAL_Conditions/Atoms/ObjectAtom.js`
- [x] `/react/components/EXPERIMENTAL_Conditions/Atoms/SubjectAtom.js`
- [x] `/react/components/EXPERIMENTAL_Conditions/Atoms/VerbAtom.js`
- [x] `/react/components/EXPERIMENTAL_Conditions/Statement.js`
- [x] `/react/components/EXPERIMENTAL_Conditions/StrategySelector.js`
- [x] `/react/components/EXPERIMENTAL_Conditions/index.js`
[...]

PR #1091, opened 2020-03-13
Fix eslint in alert component

emersonlaurentino:

#trivial Fix eslint errors #1082

Fix eslint in alert component

`yarn lint`

#### Types of changes

### DIFF
--- a/react/components/Alert/index.js
+++ b/react/components/Alert/index.js
@@ -92,7 +92,14 @@ class Alert extends Component {
             title="Close"
             className={`vtex-alert__close-icon pointer pv2 c-on-base ph3 ${innerVerticalPadding}`}
             onClick={onClose}
-            onKeyPress={onClose}
+            onKeyPress={({ key }) => {
+              const SPACE = ' '
+              const ENTER = 'Enter'
+
+              if (key === SPACE || key === ENTER) {
+                onClose()
+              }
+            }}
             role="button"
             tabIndex={0}
           >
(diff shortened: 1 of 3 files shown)


react/components/Alert/index.test.js
```diff
@@ -5,7 +5,7 @@ import Alert from './index'
 
 // Skipping those tests because getByText is getting the div inside the button and not the button element itself
 // I've tried to fix it, but I couldn't manage to. So I let me fellows future developers in this endeavour
-describe.skip('Alert', () => {
+describe('Alert', () => {
   describe('action button', () => {
     it('should be displayed if props are passed', () => {
       const label = 'Action'
```
klzns: The tests are not passing are they?

react/components/Alert/index.js
```diff
@@ -92,6 +92,8 @@ class Alert extends Component {
             title="Close"
             className={`vtex-alert__close-icon pointer pv2 c-on-base ph3 ${innerVerticalPadding}`}
             onClick={onClose}
+            onKeyPress={onClose}
```
lucasecdb: shouldn't this be checking for the type of key pressed? like an 'enter' or 'space' 🤔

lucasecdb: imagine you tabbing into this "button" and pressing tab to focus the next element and the alert closing when you didn't expect it to

react/components/Alert/index.test.js
emersonlaurentino: before they were not passed, now they are passing.

react/components/Alert/index.js
emersonlaurentino: It really must be done!

emersonlaurentino: fixed @lucasecdb

PR #1097, opened 2020-03-17
#trivial Remove unnecessary button index.js file

igorbrasileiro:

As the title says


react/components/Button/index.test.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import { render } from 'react-testing-library'
 
-import Button from './index'
+import Button from '.'
```
kaisermann:
```suggestion
import Button from './'
```

PR #1102, opened 2020-03-18
Fix eslint errors in Modal and ModalDialog

thayannevls:
#trivial Fix Eslint errors #1082

`yarn && yarn start`
`yarn lint`

#### Types of changes

### DIFF
--- a/react/components/Modal/TopBar.js
+++ b/react/components/Modal/TopBar.js
@@ -13,8 +13,11 @@ const TopBar = props => {
     showCloseIcon,
   } = props
 
-  const handleKeyDown = event => {
-    if (event.key === 'Escape' || event.key === 'Esc') {
+  const handleKeyDown = ({ key }) => {
+    const SPACE = ' '
+    const ENTER = 'Enter'
+
+    if (key === SPACE || key === ENTER) {
       onClose()
     }
   }
--- a/react/components/Modal/TopBar.js
+++ b/react/components/Modal/TopBar.js
@@ -13,6 +13,12 @@ const TopBar = props => {
     showCloseIcon,
   } = props
 
+  const handleKeyDown = event => {
+    if (event.key === 'Escape' || event.key === 'Esc') {
+      onClose()
+    }
+  }
+
   return (
     <div
       className={`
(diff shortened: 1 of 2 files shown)


kaisermann:
This PR is modifying a LOT of files unrelated to the `Modal` and `ModalDialog`. Is that correct?

react/components/ModalDialog/index.js
```diff
@@ -4,7 +4,7 @@ import PropTypes from 'prop-types'
 import Modal from '../Modal'
 import Button from '../Button'
 
-const NOOP = () => {}
+const NOOP = () => ({})
```
kaisermann: You can ignore the rule here to prevent creating an object without reason!

react/components/Modal/TopBar.js
```diff
@@ -13,6 +13,12 @@ const TopBar = props => {
     showCloseIcon,
   } = props
 
+  const handleKeyDown = event => {
+    if (event.key === 'Escape' || event.key === 'Esc') {
+      onClose()
+    }
+  }
```
lucasecdb: nice!! was about to comment that :joy:

react/components/ModalDialog/index.js
thayannevls: Done! Thanks

react/components/ModalDialog/index.js
```diff
@@ -4,6 +4,7 @@ import PropTypes from 'prop-types'
 import Modal from '../Modal'
 import Button from '../Button'
 
+// eslint-disable-next-line @typescript-eslint/no-empty-function
 const NOOP = () => {}
```
emersonlaurentino:
```suggestion
const NOOP = () => null
```

thayannevls: done :heavy_check_mark:

emersonlaurentino:
every time there is new push in PR, the approves are undone.

react/components/Modal/TopBar.js
emersonlaurentino: The correct thing is like this:

- the space key is an empty space (javascript and its jokes)

- Esc does not make sense to use, since esc has to be triggered at any time and not just when the focus is on the close button.

```suggestion
  const handleKeyDown = ({ key }) => {
    const SPACE = ' '
    const ENTER = 'Enter'

    if (key === SPACE || key === ENTER) {
      onClose()
    }
  }
```

thayannevls: True! Updated

PR #1103, opened 2020-03-18
fix eslint errors in SelectableCard, Slider and ToastProvider

thayannevls:
#trivial fix Eslint errors #1082 

`yarn & yarn start`
`yarn lint`

#### Types of changes

### DIFF
--- a/react/components/Slider/index.js
+++ b/react/components/Slider/index.js
@@ -92,10 +92,11 @@ export default class Slider extends Component {
   componentWillUnmount() {
     window.removeEventListener('resize', this.updateLayout)
 
-    if (this.cancelDragEvent_) {
-      this.cancelDragEvent_()
-      this.cancelDragEvent_ = undefined
+    if (!this.cancelDragEvent_) {
+      return
     }
+    this.cancelDragEvent_()
+    this.cancelDragEvent_ = undefined
   }
 
   updateLayout = () => {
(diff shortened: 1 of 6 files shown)


emersonlaurentino:
@thayannevls you are doing a great job, thanks for the contribution. But the next time, tell us in #1082 which component you are fixing so that we have greater control and someone does not fix the same component.

thayannevls:
Sorry @emersonlaurentino ! Next time I'll do it :)

react/components/ToastProvider/ToastManager.js
```diff
@@ -100,15 +100,15 @@ export default class ToastManager extends Component {
       windowBounds
 
     if (this.container.current) {
```
emersonlaurentino: same
```suggestion
    if (!this.container.current) {
```

react/components/Slider/index.js
```diff
@@ -93,9 +93,10 @@ export default class Slider extends Component {
     window.removeEventListener('resize', this.updateLayout)
 
     if (this.cancelDragEvent_) {
```
emersonlaurentino: the condition says that if `this.cancelDragEvent_` exists then run `this.cancelDragEvent _ ()` and `this.cancelDragEvent_ = undefined` and then return.

the way you did it, it's the other way around!

The correct should be:
```suggestion
    if (!this.cancelDragEvent_) {
```

thayannevls: My bad didn't notice it :) fixed it, thanks!

PR #1105, opened 2020-03-19
#trivial : fix inputs eslint issues

matheusps:

eslint fixes for: `Input`, `InputCurrency`, `InputPassword` & `InputSearch` #1082

#### Types of changes

### DIFF
--- a/react/components/Input/index.js
+++ b/react/components/Input/index.js
@@ -88,6 +88,10 @@ class Input extends Component {
     this.handleAutoFocus()
   }
 
+  componentDidUpdate() {
+    this.handleAutoFocus()
+  }
+
   render() {
     const {
       errorMessage,
--- a/react/components/Input/index.js
+++ b/react/components/Input/index.js
@@ -1,4 +1,4 @@
-import React, { Component } from 'react'
+import React, { Component, createRef } from 'react'
 import PropTypes from 'prop-types'
 
 import Button from '../Button'
@@ -13,6 +13,8 @@ class Input extends Component {
     this.state = {
       active: false,
     }
+
+    this.ref = createRef(null)
   }
 
   handleChange = event => {
@@ -32,17 +34,17 @@ class Input extends Component {
   }
 
   handleFocus = event => {
-    if (!this.props.readOnly) {
-      this.setState({ active: true })
-      this.props.onFocus && this.props.onFocus(event)
-    }
+    if (this.props.readOnly) return
+
+    this.setState({ active: true })
+    this.props.onFocus && this.props.onFocus(event)
   }
 
   handleBlur = event => {
-    if (!this.props.readOnly) {
-      this.setState({ active: false })
-      this.props.onBlur && this.props.onBlur(event)
-    }
+    if (this.props.readOnly) return
+
+    this.setState({ active: false })
+    this.props.onBlur && this.props.onBlur(event)
   }
 
   handleMouseEnter = event => {
@@ -51,6 +53,17 @@ class Input extends Component {
 
   handleMouseLeave = event => {
     this.props.onMouseLeave && this.props.onMouseLeave(event)
+  }
+
+  handleAutoFocus = () => {
+    const { forwardedRef, autoFocus } = this.props
+    if (!autoFocus) return
+
+    if (forwardedRef && forwardedRef.current) {
+      forwardedRef.current.focus()
+    } else if (this.ref.current) {
+      this.ref.focus()
+    }
   }
 
   componentDidMount() {
@@ -71,6 +84,8 @@ class Input extends Component {
         'You should not use both prefix and suffix props in the same input. '
       )
     }
+
+    this.handleAutoFocus()
   }
 
   render() {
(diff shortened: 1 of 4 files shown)


react/components/Input/index.js
```diff
@@ -71,6 +84,8 @@ class Input extends Component {
         'You should not use both prefix and suffix props in the same input. '
       )
     }
+
+    this.handleAutoFocus()
```
lucasecdb: shouldn't you add something like this in `componentDidUpdate` as well?

react/components/Input/index.js
```diff
@@ -208,7 +223,7 @@ class Input extends Component {
           )}
           <input
             {...dataAttrs}
-            ref={this.props.forwardedRef}
+            ref={this.props.forwardedRef || this.ref}
```
lucasecdb: can you do something like
```suggestion
            ref={node => {
              if (typeof this.props.forwardedRef === 'function') {
                this.props.forwardedRef(node)
              } else {
                this.props.forwardedRef.current = node
              }
              this.ref.current = node
            }}
```

and change the `handleAutoFocus` to only use the `this.ref`

react/components/InputSearch/index.js
```diff
@@ -73,15 +75,19 @@ class InputSearch extends Component {
         onBlur={() => this.handleFocus(false)}
         onMouseEnter={() => this.handleHovering(true)}
         onMouseLeave={() => this.handleHovering(false)}
-        onKeyUp={e => e.key === 'Enter' && this.handleSubmit(e)}
+        onKeyUp={e => this.pressedEnter(e) && this.handleSubmit(e)}
```
lucasecdb: do you know why is the `KeyUp` event being used here instead of `KeyPress`?

react/components/Input/index.js
Klynger: I think this could be done in another PR

lucasecdb: fair enough

react/components/InputPassword/index.js
```diff
@@ -18,7 +18,14 @@ class InputPassword extends Component {
...
 
-  toggle = () => this.setState(state => ({ showPassword: !state.showPassword }))
+  handleToggle = () =>
+    this.setState(state => ({ showPassword: !state.showPassword }))
+
+  handleKeyDown = event => {
+    if (event.key === 'Enter') {
+      this.handleToggle()
+    }
+  }
```
emersonlaurentino: The button has the behavior of triggering the event when pressed Space or Enter, in this case only Enter is resolved.

This solves the problem and makes reading more readable.

```suggestion
  handleKeyDown = event => {
    const SPACE = ' '
    const ENTER = 'Enter'

    if (key === SPACE || key === ENTER) {
        this.handleToggle()
    }
  }
```

react/components/InputSearch/index.js
```diff
@@ -96,7 +102,13 @@ class InputSearch extends Component {
                   InputSearch.separatorHeight.regular,
               }}
             />
-            <span className="pointer pl4 c-link" onClick={this.handleSubmit}>
+            <span
+              className="pointer pl4 c-link"
+              onClick={this.handleSubmit}
+              onKeyPress={e => this.pressedEnter(e) && this.handleSubmit(e)}
```
emersonlaurentino: you should do the same was suggested before

react/components/Input/index.js
matheusps: What do you guys think about this implementation: https://codesandbox.io/s/merging-refs-v2sry ?

react/components/Input/index.js
matheusps: yep. I'll do that.

react/components/Input/index.js
lucasecdb: perfect 👌

react/components/InputSearch/index.js
matheusps: no, I don't. I just realized that now. Don't `onKeyUp` has better performance if the user keeps the same key pressed?

react/components/Input/index.js
matheusps: Opened #1110 for that, @lucasecdb & @Klynger

react/components/InputSearch/index.js
lucasecdb: depends on the use case :sweat_smile: I was just wondering if you had an idea on why this was used instead of `onKeyPress`, don't really need to change anything

react/components/InputSearch/index.js
```diff
@@ -73,15 +84,19 @@ class InputSearch extends Component {
...
           <div className="flex flex-row items-center">
             {this.props.value && (
               <span
-                tabIndex={0}
-                onClick={this.handleClickClear}
                 className="pointer mr4 c-muted-3"
+                onClick={this.handleClickClear}
+                onKeyPress={e =>
+                  this.pressedEnter(e) && this.handleClickClear(e)
+                }
```
emersonlaurentino: here you should use `this.handleKeyPress` instead of `this.pressedEnter` no?

you can turn it into a generic function that takes a function as a parameter and then you can pass `handleSubmit` or `handleClickClear`.

matheusps: great idea. done!

PR #1106, opened 2020-03-19
#trivial: fix filterbar eslint issues

matheusps:

fixes for: `FilterBar/Menu` & `FilterBar/FilterTag` #1082

#### Types of changes

### DIFF
--- a/react/components/FilterBar/Menu.js
+++ b/react/components/FilterBar/Menu.js
@@ -100,6 +100,16 @@ class Menu extends Component {
     }
   }
 
+  handleKeyPress = event => {
+    const SPACE = ' '
+    const ENTER = 'Enter'
+    const { key } = event
+
+    if (key === SPACE || key === ENTER) {
+      this.props.onBackgroundClick(event)
+    }
+  }
+
   render() {
     const {
       width,
@@ -190,9 +200,7 @@ class Menu extends Component {
                       <div className="truncate f3 pr6">{optionsLabel}</div>
                       <div
                         onClick={this.props.onBackgroundClick}
-                        onKeyPress={e =>
-                          e.key === 'Enter' && this.props.onBackgroundClick(e)
-                        }
+                        onKeyPress={this.handleKeyPress}
                         role="button"
                         tabIndex={0}
                       >
--- a/react/components/FilterBar/Menu.js
+++ b/react/components/FilterBar/Menu.js
@@ -188,7 +188,14 @@ class Menu extends Component {
                   {isMobile && (
                     <div className="flex justify-between flex-row items-baseline pa6 mh3">
                       <div className="truncate f3 pr6">{optionsLabel}</div>
-                      <div onClick={this.props.onBackgroundClick}>
+                      <div
+                        onClick={this.props.onBackgroundClick}
+                        onKeyPress={e =>
+                          e.key === 'Enter' && this.props.onBackgroundClick(e)
+                        }
+                        role="button"
+                        tabIndex={0}
+                      >
                         <Close size={20} color="currentColor" />
                       </div>
                     </div>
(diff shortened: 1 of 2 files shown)


react/components/FilterBar/Menu.js
```diff
@@ -188,7 +188,14 @@ class Menu extends Component {
                   {isMobile && (
                     <div className="flex justify-between flex-row items-baseline pa6 mh3">
                       <div className="truncate f3 pr6">{optionsLabel}</div>
-                      <div onClick={this.props.onBackgroundClick}>
+                      <div
+                        onClick={this.props.onBackgroundClick}
+                        onKeyPress={e =>
+                          e.key === 'Enter' && this.props.onBackgroundClick(e)
+                        }
```
emersonlaurentino: The button has the behavior of triggering the event when pressed Space or Enter, in this case only Enter is resolved.

This solves the problem and makes reading more readable.

```suggestion
                        onKeyPress={({ key }) => {
                          const SPACE = ' '
                          const ENTER = 'Enter'

                          if (key === SPACE || key === ENTER) {
                            onClick()
                          }
                        }}
```

lucasecdb: @emersonlaurentino since we are doing this in a lot of components, do you think it's better to extract this in a helper function?

emersonlaurentino: I thought about it, the problem is that this is a behavior for an element that is not a button and should be. The people who developed it wrong.

estacioneto: This look like something good to be discussed there 🤔 
image

emersonlaurentino: the initial idea is to solve the eslint problems, once the problems are solved, we create the tests, then we switch to typescript and that's when we can refactor improvements. So I'm going to create an issue to make this clear and to be a place where we can write things that we would like to see done as an improvement.

emersonlaurentino: I created and pine it!

@lucasecdb comments there what you think could be an improvement for us to do, so we organize everything in one place.

@Matheusps had also commented on removing the proptypes. Comment there.

Screen Shot 2020-03-19 at 11 59 04

estacioneto: Great! So... Approved, @emersonlaurentino ? 😅

emersonlaurentino: nop, my suggestion still makes sense. What @lucasecdb commented is that my suggestion has been repeated in several other components.

estacioneto: I think your suggestion also doesn't apply to this PR scope. You are suggesting changing behavior by adding the space handling and in my opinion, is not related to `eslint`

emersonlaurentino: It actually makes sense, eslint says there is an accessibility problem when using an `onClick` in a div, and the recommendation is to use an `onKeyPress` attribute. The current behavior is not changed, it is only accessible (as eslint expects). What I am suggesting is that within the onKeyPress the correct keys are used, precisely so that the behavior is not changed.

estacioneto: I'm sorry but I still didn't get how your suggestion to add another key handling doesn't affect the behavior. If it's a bug, it should be in another PR, in my opinion. Not in this one. 
Also, wasn't bug fixes related to the `master` branch?

emersonlaurentino: I think you're confused, I'll try to explain it better.

estacioneto: As discussed in a call, It makes sense to add this accessibility behavior since you are adding the whole `onKeyPress` handler and the space handler is used in buttons, @matheusps

matheusps: Shouldn't we abstract it into a `Clickable` component? 🤔

estacioneto: It could be. Maybe an issue to discuss it would be nice

PR #1108, opened 2020-03-19
Fix the tooltip overflowing with long words

lucasaarcoverde:
As the title says

[clubhouse](https://app.clubhouse.io/vtex/story/31018/ui-implementation-resolver-problema-do-tooltip-quebrando-na-tabela-de-produtos)
[story related](https://app.clubhouse.io/vtex/story/30998/corre%C3%A7%C3%B5es-das-colunas-e-tooltips-na-tabela)

In this [workspace](https://lucasan2--cosmetics1.myvtex.com/admin/collections/) go to a collection page and on the second page of the product table you will see the example shown below

### Before
Captura de tela de 2020-03-19 11-07-55
### After
Captura de tela de 2020-03-19 11-08-12

#### Types of changes

### DIFF
--- a/react/components/Tooltip/TooltipPopup.tsx
+++ b/react/components/Tooltip/TooltipPopup.tsx
@@ -100,6 +100,7 @@ const TooltipPopup: FC<PropTypes.InferProps<typeof propTypes>> = ({
           ...positionStyle,
           zIndex: zIndex.tooltip,
           transition: `opacity ${duration}ms ${timmingFn} ${delay}ms`,
+          wordBreak: 'break-all',
         }}
         ref={popupRef}
         onTransitionEnd={() => setShowPopup(visible)}>
--- a/react/components/Tooltip/TooltipPopup.tsx
+++ b/react/components/Tooltip/TooltipPopup.tsx
@@ -44,6 +44,13 @@ const propTypes = {
   timmingFn: PropTypes.string,
   /** Child ref. Used to correctly position the tooltip */
   childRef: getChildRefPropType(),
+  /** Element that inserts line break style in the word. Used to prevent width overflow */
+  wordBreak: PropTypes.oneOf<string>([
+    'normal',
+    'break-all',
+    'keep-all',
+    'break-word',
+  ]),
 }
 
 const defaultProps = {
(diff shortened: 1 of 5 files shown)


emersonlaurentino:
@estacioneto how did you solve the similar problem with your product?

react/components/Tooltip/TooltipPopup.tsx
```diff
@@ -100,6 +100,7 @@ const TooltipPopup: FC<PropTypes.InferProps<typeof propTypes>> = ({
           ...positionStyle,
           zIndex: zIndex.tooltip,
           transition: `opacity ${duration}ms ${timmingFn} ${delay}ms`,
+          wordBreak: 'break-all',
```
estacioneto: Actually, it could be a prop. This is because the `break-all` property really breaks the word itself and, in many cases, we don't want that. It's more often to want to break in whitespace, so... Could you add a prop with it? It could be a `wordBreak` prop or even a `popupStyle` prop

estacioneto: @emersonlaurentino super large words didn't need to be handled before. I also think that changing the `max-width` could be better, but I didn't make these experiments and surely @lucasaarcoverde should have done it.

emersonlaurentino: `wordBreak` is the best solution, since we want the opinioned components.

estacioneto: I'd say static components but yeah... It can be `wordBreak` 😄

lucasaarcoverde: @estacioneto i did this test of change `max-width` and it works, but i talked to @oldufre and the best solution is using `wordBreak`. She did some research and found that is how other style guides deal with it

estacioneto: Great job! As I thought you guys would have done. But anyway can you add it as a prop instead?

lucasaarcoverde: I will change the `wordBreak` to be a prop :smile:

kaisermann:
Is there a situation where you would want to leave the default overflowing behavior?

estacioneto:
Yes, @kaisermann . In general, all words fit the `mw5` tachyons class, which is `256px` wide.

Issue #1083, opened 2020-03-12
Add test to all components


emersonlaurentino:
Add test to all components, so we can convert it to typescript.

Componentes:
- [ ] ActionMenu
- [ ] Alert
- [ ] AutocompleteInput
- [ ] Box
- [ ] Button
- [ ] ButtonGroup
- [ ] ButtonPlain
- [ ] ButtonWithIcon
- [ ] Card
- [ ] Checkbox
- [ ] CheckboxGroup
- [ ] Collapsible
- [ ] ColorPicker
- [ ] Conditions
- [ ] DatePicker
- [ ] Divider
- [ ] Dropdown
- [ ] Dropzone
- [ ] EmptyState
- [ ] EXPERIMENTAL_Charts
- [ ] EXPERIMENTAL_Conditions
- [ ] EXPERIMENTAL_Select
- [ ] EXPERIMENTAL_Table
- [ ] EXPERIMENTAL_TableTree
- [ ] EXPERIMENTAL_useCheckboxTree
- [ ] FilterBar
- [ ] FilterOptions
- [ ] FloatingActionBar
- [ ] icon
- [ ] Input
- [ ] InputButton
- [ ] InputCurrency
- [ ] InputPassword
- [ ] InputSearch
- [ ] Layout
- [ ] Link
- [ ] Menu
- [ ] Modal
- [ ] ModalDialog
- [ ] MultiSelect
- [ ] NumericStepper
- [ ] PageBlock
- [ ] PageHeader
- [ ] Pagination
- [ ] PasswordInput
- [ ] Progress
- [ ] ProgressBar
- [ ] Radio
- [ ] RadioGroup
- [ ] SelectableCard
- [ ] Slider
- [ ] Spinner
- [ ] Statement
- [ ] Tab
- [ ] Table
- [ ] Tabs
- [ ] Tag
- [ ] Textarea
- [ ] TimePicker
- [ ] ToastProvider
- [ ] Toggle
- [ ] Tooltip
- [ ] Totalizer

thayannevls:
@emersonlaurentino I will implement tests for the Modal component

igorbrasileiro:
I will implement tests for Tabs and Tab component

PR #1114, opened 2020-03-23
#trivial test: 💍 spinner component

matheusps:

As the title says

related issue #1083 

`yarn test`

#### Types of changes

### DIFF
--- a/react/components/Spinner/__tests__/index.test.js
+++ b/react/components/Spinner/__tests__/index.test.js
@@ -1,0 +1,28 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+import snapshotDiff from 'snapshot-diff'
+
+import Spinner from '../index'
+
+describe('Spinner spec', () => {
+  it('matches snapshot', () => {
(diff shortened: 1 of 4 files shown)


react/components/Spinner/__tests__/index.test.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+import snapshotDiff from 'snapshot-diff'
+
+import Spinner from '../index'
+
+describe('Spinner spec', () => {
```
thayannevls:
```suggestion
describe('Spinner', () => {
```

Shouldn't we follow a pattern for the names of the tests?

emersonlaurentino:
@matheusps can you remove the snapshot and test for the two separate colors? So we guarantee that the Spinner is actually being rendered in different colors without comparing the snapshots.

react/components/Spinner/__tests__/index.test.js
matheusps: done

PR #1115, opened 2020-03-24
Tests for the Tabs component

igorbrasileiro:
as the title says

Related issue #1083 

`yarn test`

#### Types of changes

### DIFF
--- a/react/components/Tabs/index.test.tsx
+++ b/react/components/Tabs/index.test.tsx
@@ -1,0 +1,78 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+import { toHaveClass } from '@testing-library/jest-dom/matchers'
+
+import Tabs from '.'
+import Tab from './Tab.js'
+
+describe('Tabs', () => {
+  expect.extend({ toHaveClass })
+  const tabsId = 'tabs'
+
+  describe('CSS API', () => {
+    it('props: default', () => {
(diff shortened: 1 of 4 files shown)


react/components/Tabs/index.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+      const { getByTestId } = render(
+        <Tabs testId={tabsId}>
+          <Tab label="label 1" active>
+            Tab 1
+          </Tab>
+          <Tab label="label 2">Tab 2</Tab>
+        </Tabs>
+      )
+
+      expect(getByTestId(tabsId)).toHaveTextContent('Tab 1')
```
klzns: [Avoid using getByTestId when possible](https://testing-library.com/docs/guide-which-query)

```suggestion
      expect(getByText('Tab 1')).toBeTruthy()
```

igorbrasileiro: Nice @klzns, I will rethink about the way I did my tests.

emersonlaurentino: but you don't need to remove the `testId` prop from your component, test only.

igorbrasileiro: I removed from tab because if I want to get it, I can use getByText that's suggested by `Testing Library`

PR #1117, opened 2020-03-24
Fix AutocompleteInput styles

rerissondaniel:

As the title says

Type `veg` in the products autocomplete input
[workspace](https://fixautocompletetext--cosmetics1.myvtex.com/admin/app/collections/2340/)

Before
chrome-capture (1)
After
chrome-capture (2)

#### Types of changes

### DIFF
--- a/react/components/AutocompleteInput/Option.tsx
+++ b/react/components/AutocompleteInput/Option.tsx
@@ -54,11 +54,11 @@ const Option: React.FunctionComponent<PropTypes.InferProps<
     const match = value.substr(index, searchTerm.length)
     const suffix = value.substring(index + match.length)
     return (
-      <span className="truncate">
-        <span className="fw7">{prefix}</span>
-        {match}
-        <span className="fw7">{suffix}</span>
-      </span>
+      <>
+        {prefix}
+        <span className="fw7">{match}</span>
+        {suffix}
+      </>
     )
   }
 
@@ -75,9 +75,9 @@ const Option: React.FunctionComponent<PropTypes.InferProps<
       onMouseEnter={() => setHighlightOption(true)}
       onMouseLeave={() => setHighlightOption(false)}
       onClick={onClick}>
-      <span className="h1 flex items-center">
+      <span className="h1 flex items-center truncate">
         <span className="mr3 c-muted-2 flex pt1">{icon}</span>
-        {renderOptionValue()}
+        <span className="truncate">{renderOptionValue()}</span>
       </span>
     </button>
   )
--- a/react/components/AutocompleteInput/Option.tsx
+++ b/react/components/AutocompleteInput/Option.tsx
@@ -55,9 +55,9 @@ const Option: React.FunctionComponent<PropTypes.InferProps<
     const suffix = value.substring(index + match.length)
     return (
       <>
-        {prefix}
-        <span className="fw7">{match}</span>
-        {suffix}
+        <span className="fw7">{prefix}</span>
+        {match}
+        <span className="fw7">{suffix}</span>
       </>
     )
   }
--- a/react/components/AutocompleteInput/Option.tsx
+++ b/react/components/AutocompleteInput/Option.tsx
@@ -75,7 +75,7 @@ const Option: React.FunctionComponent<PropTypes.InferProps<
       onMouseEnter={() => setHighlightOption(true)}
       onMouseLeave={() => setHighlightOption(false)}
       onClick={onClick}>
-      <span className="h1 flex items-center truncate">
+      <span className="h1 flex items-center">
         <span className="mr3 c-muted-2 flex pt1">{icon}</span>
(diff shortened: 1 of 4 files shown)


react/components/AutocompleteInput/Option.tsx
```diff
@@ -54,11 +54,11 @@ const Option: React.FunctionComponent<PropTypes.InferProps<
...
-      <span className="truncate">
-        <span className="fw7">{prefix}</span>
-        {match}
-        <span className="fw7">{suffix}</span>
-      </span>
+      <>
+        {prefix}
+        <span className="fw7">{match}</span>
+        {suffix}
+      </>
```
estacioneto: How was this bold text change validated?

react/components/AutocompleteInput/Option.tsx
```diff
@@ -75,9 +75,9 @@ const Option: React.FunctionComponent<PropTypes.InferProps<
       onMouseEnter={() => setHighlightOption(true)}
       onMouseLeave={() => setHighlightOption(false)}
       onClick={onClick}>
-      <span className="h1 flex items-center">
+      <span className="h1 flex items-center truncate">
```
estacioneto: Why is it necessary here?

react/components/AutocompleteInput/Option.tsx
rerissondaniel: I'll wait for oldufre's approval, but I think that doesnt make sense to highlight the text that doesn't match with the user input

estacioneto: Well, I've done it this way because we've decided it together. A pretty good example:

image

react/components/AutocompleteInput/Option.tsx
rerissondaniel: It wasn't, thanks

react/components/AutocompleteInput/Option.tsx
estacioneto: The autocomplete idea is not to highlight what matches. Its idea is to highlight the **suggestions**. Your logic makes perfect sense in the `Select` component.

matheusps: I agree with @estacioneto

rerissondaniel: Oh, ok then. But in that case we only have suffixes, then its a different case, no?

estacioneto: Actually not

image

estacioneto: [Kinda OFF] Related to the `admin-collections`... Did you guys notice that your autocomplete filter is not working?

rerissondaniel: Ok, I've reverted the bold

estacioneto:
It feels a little bit weird to me that suggestions are truncated. You cannot see the whole suggestion... For example (Google again):

image

I really didn't think in a better way to solve this to **huge/massive** single words, which looks like to be the case. So @oldufre really agreed with it, right?

react/components/AutocompleteInput/Option.tsx
rerissondaniel: I still think is weird because most of suggestions in these cases are in the middle, but it may be just effect of the products names... About the filter I've managet to filter after pressing enter, but maybe would be good if the results were filtered as the user types

estacioneto: Totally. The filter cannot happen when pressing enter. This behavior should be related to the user's ability search by term.

rerissondaniel: But anyway, I still feel weird in our case because is not unusual to have options that don't have a matching text in results while we're waiting for results... But I don't know, may be just a biased perception

estacioneto:
@rerissondaniel I saw that what you've done was fix when there is no match. Truncate is how we were dealing with it anyway. Just fix your CHANGELOG.md message then (`Truncate` instead of `truncate`).

react/components/AutocompleteInput/Option.tsx
estacioneto: You can filter the current results when typing. I remember we just did it that way because it's a paged resource and we didn't want to add another filter logic when, in some moment, the GraphQL API would do it.

rerissondaniel:
BTW, the semantics is different depending on the suggestion in chrome search bar:
Captura de tela de 2020-03-24 15-18-54

estacioneto:
Cool, @rerissondaniel ! It really mixes the autocomplete with a `Select`

rerissondaniel:
Yes, and as our API returns the full name of products instead of search completions we have an autocomplete that consumes Select data!

estacioneto:
About full names:

IMG_20200325_093850

Anyway you are right in collections case. The semantic is a little bit different but your questions make sense. But it's not a Styleguide problem to deal.

PR #1118, opened 2020-03-25
#trivial fix button and button tests eslint issues

igorbrasileiro:
as the title says. #1082  

#### Types of changes

### DIFF
--- a/react/components/Button/index.tsx
+++ b/react/components/Button/index.tsx
@@ -27,7 +27,7 @@ const propTypes = {
   /** Loading state */
   isLoading: PropTypes.bool,
   /** [DEPRECATED] If you are using just an Icon component inside, use this as true */
-  icon: PropTypes.bool,
+  icon: PropTypes.bool.isRequired,
   /** @ignore For internal use
    * Sets reduced paddings in order to keep the button squareish if it
    * only has an icon  */
@@ -41,7 +41,7 @@ const propTypes = {
   /** (Button spec attribute) */
   autoComplete: PropTypes.string,
   /** (Button spec attribute) */
-  disabled: PropTypes.bool,
+  disabled: PropTypes.bool.isRequired,
   /** @ignore Forwarded Ref */
   forwardedRef: refShape,
   /** (Button spec attribute) */
--- a/react/components/Button/index.tsx
+++ b/react/components/Button/index.tsx
@@ -27,7 +27,7 @@ const propTypes = {
   /** Loading state */
   isLoading: PropTypes.bool,
   /** [DEPRECATED] If you are using just an Icon component inside, use this as true */
-  icon: PropTypes.bool.isRequired,
+  icon: PropTypes.bool,
   /** @ignore For internal use
    * Sets reduced paddings in order to keep the button squareish if it
    * only has an icon  */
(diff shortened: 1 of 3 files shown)


react/components/Button/index.tsx
```diff
@@ -41,7 +41,7 @@ const propTypes = {
   /** (Button spec attribute) */
   autoComplete: PropTypes.string,
   /** (Button spec attribute) */
-  disabled: PropTypes.bool,
+  disabled: PropTypes.bool.isRequired,
```
emersonlaurentino: when they use the button, is it necessary to change it to prop disabled?

react/components/Button/index.tsx
```diff
@@ -27,7 +27,7 @@ const propTypes = {
   /** Loading state */
   isLoading: PropTypes.bool,
   /** [DEPRECATED] If you are using just an Icon component inside, use this as true */
-  icon: PropTypes.bool,
+  icon: PropTypes.bool.isRequired,
```
emersonlaurentino: same with icon?

react/components/Button/index.tsx
igorbrasileiro: Nope, but disabled and icon is by default props false. If this props isn't required, eslint will warn to fix this [line](https://github.com/vtex/styleguide/blob/375d01193da371167c2e5a861024bdab43eb2f6f/react/components/Button/index.tsx#L162). The eslint fix changes from `||` to `??` operators, but these two operators aren't equivalents.

emersonlaurentino: but it is not obliging the user to pass the `disabled` and `icon` props to the button that will resolve this.

igorbrasileiro: Ok, so can I remove it from default props?

react/components/Button/index.tsx
```diff
@@ -159,8 +159,8 @@ const Button: FC<Props> = props => {
     }
   }
 
-  const disabled = props.disabled || isLoading
-  const iconOnly = icon || props.iconOnly
+  const disabled = (props.disabled as boolean) || isLoading
+  const iconOnly = (icon as boolean) || props.iconOnly
```
emersonlaurentino:
```suggestion
  const disabled = !!props.disabled || isLoading
  const iconOnly = !!icon || props.iconOnly
```

igorbrasileiro: fixed

react/components/Button/index.tsx
```diff
@@ -159,7 +159,9 @@ const Button: FC<Props> = props => {
     }
   }
 
+  // eslint-disable-next-line
   const disabled = props.disabled || isLoading
+  // eslint-disable-next-line
```
emersonlaurentino, line 164:
```suggestion
  // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
```

react/components/Button/index.tsx
```diff
@@ -159,7 +159,9 @@ const Button: FC<Props> = props => {
     }
   }
 
+  // eslint-disable-next-line
```
emersonlaurentino, line 162:
```suggestion
  // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
```

react/components/Button/index.tsx
```diff
@@ -159,7 +159,9 @@ const Button: FC<Props> = props => {
     }
   }
 
+  // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
```
matheusps, line 162: don't disable

emersonlaurentino: why?

Issue #1131, opened 2020-04-01
First Dropdown's option keep selected when set value to undefined


ThaynanAndrey:
**Describe the bug**
When I select undefined as the value of Dropdown, shows the first option on Dropdown as selected.

**To Reproduce**

1. Render Dropdown with value other than undefined;
2. Set value to undefined;
3. Open Dropdown's options;
4. Observe that firt option is selected.

**Expected behavior**
No option should be selected

emersonlaurentino:
fixed in #1132

PR #1132, opened 2020-04-01
Fixed Dropdown's selected option when value is undefined

ThaynanAndrey:
As title says.

Described at #1131 

#### Types of changes


emersonlaurentino:
@thaynamatos for some reason is not working, I tested it locally and when the `onChange` event is triggered... the Component re-renders +5x making the result of `value === option.value` to be `false`.

PR #1136, opened 2020-04-02
#trivial Fix feature/typescript-responsive-tabs hooks issues

matheusps:

💡Refactor Tab to typescript
💡Created util file
🚀Created `useDevice` hook to avoid that avoids re-renders
🚀Improve `calculateTabsVisibility` logic to avoid `maximum deep exceeded` problems.
💄Changed `handleShowSelectedHiddenTab` to  be pure
💄Changed `handleHideTabs` to deal with undefined/null values
💄Removed eslint disables


react/components/utils/useDevice.tsx
```diff
@@ -0,0 +1,30 @@
+import { useMemo } from 'react'
```
igorbrasileiro, line 1: can use useMedia directly

react/components/Tabs/util.ts
```diff
@@ -0,0 +1,91 @@
...
+  hideTabs: boolean
+  tabIndex: number
+}
+
+interface HandleShowSelectedHiddenTabInput extends HandleHideTabsOutput {
+  tabsOrderList: number[]
+  selectedTabIndex: number
+}
+
+export const mapArrayToIndex: <T>(array: T[]) => number[] = array =>
```
igorbrasileiro, line 23: if won't use remove it

react/components/Tabs/index.tsx
```diff
@@ -139,62 +48,58 @@ const Tabs: FC<InferProps<typeof propTypes>> = ({
...
       fullWidth,
       isMobile,
     })
 
-    // change display tabs order - every hidden selected tab should be displayed
-    const newTabsOrderList = mapArrayToIndex(childrenArray)
-    handleShowSelectedHiddenTab({
-      tabsOrderList: newTabsOrderList,
+    const newOrderList = handleShowSelectedHiddenTab({
+      tabsOrderList: childrenArray.map((_, index) => index),
```
igorbrasileiro, line 81: use mapArrayToIndex or remove

react/components/Tabs/index.tsx
```diff
@@ -139,62 +48,58 @@ const Tabs: FC<InferProps<typeof propTypes>> = ({
...
+    const tabs = tabsContainerRef?.current?.children
 
     // verify if is necessary hide tabs
     const { hideTabs, tabIndex } = handleHideTabs({
       tabsContainerWidth,
       tabsContainerFullWidth: tabsFullContainerRef.current?.clientWidth,
       tabs,
       selectedTabIndex,
-      tabsOrderList,
+      tabsOrderList: childrenArray.map((_, index) => index),
```
igorbrasileiro, line 75: use mapArrayToIndex or remove

react/components/Tabs/index.tsx
```diff
@@ -139,62 +48,58 @@ const Tabs: FC<InferProps<typeof propTypes>> = ({
   const tabsFullContainerRef = useRef<HTMLDivElement>(null)
   // Handle tabs menu actions
   const moreTabsButtonRef = useRef<HTMLButtonElement>(null)
-  const tabsMenuRef = useRef<Menu>(null)
+  const tabsMenuRef = useRef<HTMLElement>(null)
 
   const selectedTabIndex: number = childrenArray.reduce(
```
igorbrasileiro, line 53: useMemo here to not recalculate

react/components/Tabs/index.tsx
```diff
@@ -1,130 +1,39 @@
...
-
-const Tabs: FC<InferProps<typeof propTypes>> = ({
-  children,
-  fullWidth,
-  isMobile,
-  sticky,
-}) => {
-  const childrenArray: Tab[] = Children.toArray(children)
+function Tabs({ children, fullWidth = false, sticky = false }: Props) {
+  const childrenArray = React.useMemo(
```
igorbrasileiro, line 32: children change every render, this may cause an overhead instead of help.

react/components/Tabs/index.tsx
```diff
@@ -1,130 +1,39 @@
...
 import debounce from 'lodash/debounce'
 
-import Tab from './Tab'
+import { TabProps } from './Tab'
 import Menu from '../Menu'
 import OptionsDots from '../icon/OptionsDots'
-import withDevice from '../utils/withDeviceHoc'
+import useDevice from '../utils/useDevice'
+import {
+  mapArrayToIndex,
```
igorbrasileiro, line 19: remove if won't use

react/components/Tabs/index.tsx
matheusps: done

react/components/Tabs/index.tsx
matheusps: not really. This is being used as a dependency on some effects. Without `useMemo` the effects will be called forever.

react/components/utils/useDevice.tsx
matheusps: The behaviors are reused in both tabs and tab

igorbrasileiro: can use in both

matheusps: but, I cannot reuse the conditions and the two hook calls

matheusps: I would literary copy & paste the hook body on each component, instead of just `const { isMobile } = useDevice()`. (the same  for other components in the styleguide that may/will need this in the future)

igorbrasileiro: you just need const isMobile = useMedia({ maxWidth: '40rem' }, true)

PR #1151, opened 2020-04-07
Add 'fit' prop to PageBlock and Box components

### LINKED ISSUE #1147: PageBlock and Box components should have "stretch" variation
**Describe the bug**

We have no way of making a `PageBlock` component fill its parent's height.
Also, shouldn't the `aside` variation be able to stretch blocks and make them the same height?

**To Reproduce**

Just create two `PageBlock` components, side by side, and make one of them be bigger than the other one. E.g.:

image

**Expected behavior**

image

estacioneto:

As the title says... Fixes #1147 

[Running workspace](https://refact--madeirasgasometro.myvtex.com/admin/app/vtexlog/reports/performance?month=1&quarter=false). You can see the boxes have the same height.

##### Before

image

##### After

image

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [9.113.0] - 2020-04-07
+
 ### Added
 
 - `fit` prop to `PageBlock` and `Box` components.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 
-- `fill` prop to `PageBlock` and `Box` components.
+- `fit` prop to `PageBlock` and `Box` components.
 
 ## [9.112.28] - 2020-04-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- `fill` prop to `PageBlock` and `Box` components.
+
 ## [9.112.28] - 2020-04-02
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- `fill` prop to `PageBlock` and `Box` components.
```
Klynger, line 12: It could be a prop named `displayMode` or something and accept the values `['fill', 'default']` and then you could add some other options that just add `w-100` or just add `h-100`

estacioneto: It can be done this way. I don't see much value in making the `w-100` and `h-100` separately because the `w-100` is already the default behavior (I just added in this PR to be explicit but it can be removed). Anyway, `displayMode` feels weird. Do you have a better idea?

Klynger: `stretchMode` I think, but the idea is to create a prop that is an enum instead of creating a boolean

estacioneto: @matheusps @emersonlaurentino any suggestions?

emersonlaurentino: I was just looking at a name for the prop. The idea is precisely to reduce the use of boolean props.

emersonlaurentino: I liked the suggestion of @Klynger

Klynger: Which one? @emersonlaurentino

emersonlaurentino: `displayMode`

emersonlaurentino: but did we get a more specific name?

emersonlaurentino: https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit

estacioneto: This fit may not be the one because only makes sense to us the `fill` and `cover` options, I think...

estacioneto: But if we agree in a good name for this prop, I can make the change with no problem

emersonlaurentino: in object fit there are these two options. But I don't like that name...

Klynger: Maybe `fit: 'fill' | 'fill-horizontal' | 'fill-vertical' | 'none'`? Or something com fit, like `fitMode`

estacioneto: Okay... I'll do it.

PR #1160, opened 2020-04-09
Fix eslint in core files

emersonlaurentino:

#trivial Fix eslint and typescript errors #1082

Fix eslint in core files and remove files that is not used anywhere

`yarn lint`

#### Types of changes


react/modules/deprecated.tsx
```diff
@@ -14,19 +14,12 @@
...
-interface UseNewProps {
-  submitFilterLable: string
-  newFilterLable: string
-}
-
-interface UseNewComponent {
-  old: string
-  new: string
-}
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
matheusps, line 17: should we disable rules?

emersonlaurentino: in this case yes, because there is no way to know which type of prop is going to be passed.

react/modules/deprecated.tsx
```diff
@@ -14,19 +14,12 @@
...
+type genericObj = Record<string, any>
```
arthurepc, line 18: ⛏️ This use of `Record` with any is a bit strange to me, I would go for: 
```suggestion
interface genericObj {
    [key: string]: any
}
```
I don't know if this works the same as the line above. What do you think?

emersonlaurentino: it is exactly the same thing, for those who are used to typescript Record it is much simpler to read. First parameter is the key and the second is the value, always.

PR #1167, opened 2020-04-13
[Conditions] Fix lint errors

igorbrasileiro:
As the title says. #trivial #1083 

#### Types of changes

### DIFF
--- a/react/components/Statement/index.tsx
+++ b/react/components/Statement/index.tsx
@@ -1,38 +1,49 @@
 import React from 'react'
 
 import SubjectAtom, { SubjectOptions } from './Atoms/SubjectAtom'
-import VerbAtom from './Atoms/VerbAtom'
+import VerbAtom, { VerbOption } from './Atoms/VerbAtom'
 import ObjectAtom from './Atoms/ObjectAtom'
+import { StatementProp } from './typings'
 
-type Props = {
+type StatementProps = {
   isFullWidth?: boolean
   isRtl?: boolean
   omitSubject?: boolean
   omitVerbs?: boolean
-  onChangeStatement: (statement: Props['statement']) => void
+  onChangeStatement: (statement: StatementProp) => void
   options: SubjectOptions
-  statement?: {
-    subject: string
-    verb: string
-    object?: unknown
-    error?: string
-  }
+  statement?: StatementProp
   subjectPlaceholder: string
 }
 
-const Statement: React.FC<Props> = ({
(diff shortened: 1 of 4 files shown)


react/components/Statement/index.tsx
```diff
@@ -76,7 +77,7 @@ const Statement: React.FC<Props> = ({
       disabled={!statement.verb}
       error={statement.error}
       object={statement.object}
-      onChange={(object, error = null) => {
+      onChange={(object, error = undefined) => {
```
kaisermann, line 80: If no value is passed, the default value would be `undefined`, no?

react/components/Statement/index.tsx
```diff
@@ -3,36 +3,39 @@ import React from 'react'
...
-    statement.subject &&
-    options[statement.subject].verbs.find(verb => verb.value === statement.verb)
+  const verbOptions = (Boolean(statement.subject) &&
+    options[statement.subject].verbs.find(
+      verb => verb.value === statement.verb
+    )) || {
+    label: '',
+    value: '',
+    object: NoopComponent,
+  }
```
kaisermann, line 38: Boy what is happening here 😆 

I'd rewrite this to something a bit more readable, like:

```suggestion
let verbOptions = {
	label: '',
	value: '',
	object: NoopComponent,
};

if (statement.subject) {
	const foundOption = options[statement.subject].verbs.find(
		(verb) => verb.value === statement.verb,
	);

	if (foundOption) {
		verbOptions = foundOption;
	}
}
```

igorbrasileiro: Fixed

react/components/Statement/index.tsx
```diff
@@ -1,38 +1,49 @@
...
+  statement = { subject: '', verb: '', object: null },
   subjectPlaceholder,
 }) => {
-  const verbOptions =
-    statement.subject &&
-    options[statement.subject].verbs.find(verb => verb.value === statement.verb)
+  let verbOptions: VerbOption = {
+    label: '',
+    value: '',
+    object: NoopComponent,
```
emersonlaurentino, line 34: couldn't it be a 'null' here?

igorbrasileiro: Nope, because object is type `ObjectOption = (renderProps: RenderProps) => React.ReactElement`. Only if i change this `ObjectOption` type, but i think it's not a good idea.

Issue #1146, opened 2020-04-03
Checkbox prop "partial" isn't HTML compliant


lucasecdb:
**Describe the bug**
The prop `partial` in the `Checkbox` component doesn't respect HTML's specification.

**To Reproduce**

1. Go to [Styleguide Checkbox's docs](https://styleguide.vtex.com/#/Components/Forms/Checkbox)
2. Inspect the "partial" checkbox and store the input element reference to use in the console.
3. See the value of the property `indeterminate` of the input element.

**Expected behavior**

To have the indeterminate property of the `input` element set to `true`.

**Screenshots**

<img width="426" alt="image" src="https://user-images.githubusercontent.com/10223856/78368885-b75d9b80-759a-11ea-8acc-0f370e1ea867.png">

The `temp0` variable is a reference to the second checkbox `input`'s.

**Additional context**

[Documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#Indeterminate_state_checkboxes).

lucasecdb:
can we also rename this prop to `indeterminate`? so it's more "aligned" with the HTML spec

PR #1169, opened 2020-04-13
fix: 🐛 using indeterminate on checkbox when partial true

emersonlaurentino:

add value to indeterminate attribute

✅ Closes: #1146

add this code on sandbox in the styleguide docs site

```js
class CheckboxWithRef extends React.Component {
  constructor() {
    super(); 
    this.state = { checked: false };
    this.reference = React.createRef();
  }

  render() {
    window.REFC = this.reference

    return (
    <Checkbox
      checked={this.state.checked}
      partial
      id="option-partial"
      label="Partial"
      ref={this.reference}
      name="default-checkbox-group"
      onChange={e => this.setState({ checked: !this.state.checked })}
      value="option-partial"
    />
    )
  }
}

;<CheckboxWithRef />
```

You can find element like this:

```js
t = document.querySelector("#option-partial")
t.indeterminate
// true
```

or using ref

```js
window.REFC.current.indeterminate
// true
```

Screen Shot 2020-04-13 at 18 59 25

#### Types of changes

### DIFF
--- a/react/components/Checkbox/index.js
+++ b/react/components/Checkbox/index.js
@@ -7,6 +7,24 @@ import CheckPartial from '../icon/CheckPartial'
 import { withForwardedRef, refShape } from '../../modules/withForwardedRef'
 
 class Checkbox extends PureComponent {
+  constructor(props) {
+    super(props)
+    this.myCheckbox = React.createRef()
+  }
+
+  componentDidMount() {
+    this.myCheckbox.indeterminate = !this.props.checked && this.props.partial
+  }
+
+  componentDidUpdate(prevProps) {
+    if (
+      prevProps.partial !== this.props.partial ||
+      prevProps.checked !== this.props.checked
+    ) {
+      this.myCheckbox.indeterminate = !this.props.checked && this.props.partial
+    }
+  }
+
   handleChange = e =>
     this.props.onChange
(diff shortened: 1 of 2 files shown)


react/components/Checkbox/index.js
```diff
@@ -79,7 +79,10 @@ class Checkbox extends PureComponent {
           </div>
           <input
             checked={checked}
-            ref={forwardedRef}
+            ref={elem => {
+              elem && (elem.indeterminate = !checked && partial)
```
lucasecdb, line 83: can we do this in `componentDidMount`/`componentDidUpdate`? feels a bit hacky to do this in the ref callback, which I don't think is supposed to mutate the DOM node

emersonlaurentino: there is no difference beyond the number of lines of code

estacioneto: I agree with @lucasecdb ... I have never seen someone mutate a DOM node in the ref callback. It really looks like it's a job for the callbacks he mentioned.

emersonlaurentino: the problem is not the mutation, because regardless of where it is being done (inline or in its life cycle) it will be a mutation. According to gaearon the problem inline is apparently related to server rendering. So i will change to use DidMount and DidUpdate!

PR #1177, opened 2020-04-27
New Modal

thayannevls:
- Rewrite Modal and ModalDialog
- Improve component accessibility.
- Add Sizes.
- Add `aria-label`, `aria-labelled-by` and `aria-describedby`. 
- Fix `centered`.
- Fix `ModalDialog` responsiveness on Mobiles.
- Change TopBar and BottomBar paddings.
- Improve the close button style.
- Allow `div` props into the component.
- Add news examples in docs and storybook
- Improve tests
- Add `data-testid`

Modal with sizes is a current request at Styleguide, also our old Modal had some bugs that couldn't be fixed using the `react-responsive-modal`.

[Storybook](http://b20d3072.ngrok.io/?path=/story/components-modal--default), If the link is down contact me at slack @ thayanne, or ping me here

Please, if you are a dev prefer to run storybook and styleguide local :pleading_face: 
`yarn && yarn storybook`
`yarn && yarn styleguide`

#### Types of changes


react/components/Modal/TopBar.tsx
```diff
@@ -0,0 +1,53 @@
...
+    if (key === SPACE || key === ENTER) {
+      onClose()
+    }
+  }
+
+  return (
+    <div className="mb5">
+      <div className="min-h-large pl8">
+        {showCloseIcon && (
+          <div
```
matheusps, line 31: You can use a `button` instead of faking the behavior of one. You may take a look at the `ButtonPlain` component as well.

react/components/Modal/TopBar.tsx
```diff
@@ -0,0 +1,53 @@
...
+            onClick={onClose}
+            ref={forwardedRef}
+            tabIndex={0}
+          >
+            <IconClose color="black" />
+          </div>
+        )}
+      </div>
+      {showTopBar && children && (
+        <span className="t-heading-4 ml8">{children}</span>
```
matheusps, line 47:
```suggestion
        <h4 className="t-heading-4 ml8">{children}</h4>
```

react/components/Modal/index.tsx
```diff
@@ -0,0 +1,281 @@
...
+          {
+            'vw-40-ns': size === 'small',
+            'vw-50-ns': size === 'medium',
+            'vw-60-ns': size === 'large',
+            'h-100 h-auto-ns vw-100': responsiveFullScreen,
+            'vw-50': !responsiveFullScreen,
+          }
+        )}
+        onClick={e => e.stopPropagation()}
+        role="dialog"
```
matheusps, line 141:
```suggestion
        role="dialog"
        aria-modal="true"
```
Also missing aria-labelledby & aria-describedby

react/components/Modal/index.tsx
```diff
@@ -0,0 +1,281 @@
...
+          showCloseIcon={showCloseIcon}
+          onClose={onClose}
+          showTopBar={showTopBar}
+          responsiveFullScreen={responsiveFullScreen}
+        >
+          {title}
+        </TopBar>
+        <div
+          className={`ph8 t-body overflow-auto flex flex-column flex-shrink-1 flex-grow-1 mb3 ${styles.scrollBar}`}
+          style={{ maxHeight: '60vh' }}
```
matheusps, line 155: You can create a token for this in your CSS since you have one.

react/components/Modal/modal.stories.tsx
```diff
@@ -0,0 +1,166 @@
...
+      <Button onClick={open} type="button">
+        Open
+      </Button>
+      <Modal
+        isOpen={isOpen}
+        onClose={close}
+        size={select('Size', sizes, 'medium')}
+        responsiveFullScreen
+      >
+        {lorem}
```
matheusps, line 81: There are some great texts on the current Modal & ModalDialog docs. You could reuse them.

react/components/Modal/TopBar.tsx
thayannevls: I can't get the same appearance with these components because of their hover background :/

react/components/Modal/TopBar.tsx
thayannevls: `h4` element adds the `display: block` property which is not desirable here

react/components/Modal/index.tsx
thayannevls: :thinking: True, since I can't add anything to these properties because they must be descriptive about the Modal content I will add them as properties in the Modal

react/components/Modal/modal.stories.tsx
thayannevls: Yeah, I saw it but I didn't add here because I write all the stories in one file and big texts make the file really extensive and not easy to read

react/components/Modal/index.tsx
thayannevls: Didn't notice this one, thanks!

thayannevls:
Keep in mind that maybe I'll change something about the sizes with @oldufre in the next days :smile:

react/components/Modal/BottomBar.tsx
```diff
@@ -0,0 +1,32 @@
...
+        'flex justify-content flex-row-reverse min-h-regular-ns min-h-small pv6-ns ph8-ns pv7 ph6 mt3',
+        {
+          'bt b--muted-4': showBorder,
+          pb7: responsiveFullScreen,
+        }
+      )}
+      style={responsiveFullScreen ? { marginTop: 'auto' } : {}}
+      ref={forwardedRef}
+    >
+      <div>{children}</div>
```
emersonlaurentino, line 27: Do you need this `div`?
```suggestion
      {children}
```

react/components/Modal/useModal.ts
```diff
@@ -0,0 +1,10 @@
+import { useState } from 'react'
+
+const useModal = () => {
```
emersonlaurentino, line 3: I think we could rename this hook to something like `useDisclosure` just like [ChackraUI](https://chakra-ui.com/usedisclosure) does.

I also think that it should be a separate component and not a part of the modal. Honestly I don't like none `useDisclosure`, we could see other suggestions.

emersonlaurentino:
@thayannevls I did not continue the review, because I believe that as there will be many modifications with respect to `useModal`, it does not make sense for me to continue now.

react/components/Modal/BottomBar.tsx
thayannevls: Yes, because I use `flex-row-reverse` in the div above to push elements to the right but I don't this property to change the children

react/components/Modal/useModal.ts
thayannevls: I agree we can rename it. I created this hook because I noticed that everyone uses the Modal in the same way that the hook is defined

react/components/Modal/index.tsx
```diff
@@ -0,0 +1,288 @@
...
+          className={classNames(
+            'flex fixed z-max overflow-hidden overflow-hidden bg-black-70 absolute--fill',
+            isOpen ? styles.openAnimation : styles.closeAnimation,
+            {
+              'items-start': !centered,
+              'items-center': centered,
+            }
+          )}
+          tabIndex={-1}
+          onClick={handleClick}
```
Klynger, line 96:
```suggestion
          onClick={handleClick}
          role="presentation"
```

react/package.json
```diff
@@ -5,10 +5,10 @@
     "react-color": "^2.17.0",
     "react-datepicker": "^2.0.0",
     "react-dropzone": "^10.1.8",
+    "react-focus-lock": "^2.3.1",
```
Klynger, line 8: This lib is a [bit big for what it does](https://bundlephobia.com/result?p=react-focus-lock@2.3.1), maybe you can implement it. [Here is an example](https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/Modal/TrapFocus.js)

react/components/Modal/TopBar.tsx
thayannevls: Done!

react/components/Modal/modal.css
```diff
@@ -26,3 +18,60 @@
...
+  .mw40 {
+    max-width: 40rem;
+  }
+}
+
+@media screen and (min-width: 40em) {
+  .mw50 {
+    max-width: 50rem;
+  }
+}
```
Klynger, line 51: I don't think we should create classes like tachyons, because if we decide to add some handle like this in the tachyons this one will have more priority

Klynger: And I think you should try to use tachyons instead of writing these values

react/components/Modal/modal.css
```diff
@@ -26,3 +18,60 @@
...
+    opacity: 0;
+  }
+}
+
+.openAnimation {
+  animation: fade-in 500ms;
+}
+
+.closeAnimation {
+  animation: fade-out 500ms;
```
Klynger, line 76: This animation seems to long, but I don't know

react/components/ModalDialog/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+    )
+  }, [])
+
+  const handleCancelation = () => {
+    cancelation.onClick?.()
+  }
+
+  const handleConfirmation = () => {
+    confirmation.onClick?.()
+  }
```
Klynger, line 40: I think you don't need to create these function, you could just pass it to the elements

react/components/Modal/modal.css
thayannevls: It's the same time that had before, I didn't change it

react/components/Modal/modal.css
thayannevls: Tachyons don't have `max-width`, but I will change the names as you suggested

Klynger: It does https://github.com/vtex/vtex-tachyons/blob/master/tachyons.css#L1370

Klynger: > Tachyons don't have `max-width`, but I will change the names as you suggested

It does https://github.com/vtex/vtex-tachyons/blob/master/tachyons.css#L1370 @thayannevls

thayannevls: Done! 

http://5d6fca40.ngrok.io/?path=/story/components-modal--default

thayannevls: Sorry @Klynger , I was trying to say that tachyons don't have the size that the designer requested for the Modal sizes :/

Klynger: Yeah, but in this case shouldn't the designer reconsider maybe? Because those sizes seem like customizations of the app that the user would do

thayannevls: I'll discuss with @oldufre about this but if we take longer to decide about this, do you think it's okay to merge with these classes and maybe change later?

Klynger: The store isn't using this component (because if it was, this would be a breaking change), so I think it's ok

react/components/ModalDialog/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+        <div>
+          <span className="mr4">
+            <Button
+              size="small"
+              type="button"
+              variation="tertiary"
+              disabled={loading}
+              onClick={() => {
+                cancelation.onClick?.()
+              }}
```
Klynger, line 47:
```suggestion
              onClick={cancelation.onClick}
```

react/components/ModalDialog/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+            </Button>
+          </span>
+          <Button
+            size="small"
+            type="button"
+            variation={confirmation.isDangerous ? 'danger' : 'primary'}
+            isLoading={loading}
+            onClick={() => {
+              confirmation.onClick?.()
+            }}
```
Klynger, line 59:
```suggestion
            onClick={confirmation.onClick}
```

react/components/ModalDialog/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+  label: string
+  onClick: () => unknown
+}
+
+type Props = Pick<
+  ModalProps,
+  'children' | 'isOpen' | 'onClose' | 'onCloseTransitionFinish' | 'container'
+> & {
+  loading?: boolean
+  confirmation: DialogOption & { isDangerous?: boolean | null }
```
Klynger, line 17: Sorry, didn't see this before, why not instead of using a boolean receive just the `variation` from the button? Or if you really want to restrict the options here you could
```suggestion
  confirmation: DialogOption & { variation?: 'primary' | 'danger' }
```

This way we won't have to create something like `isSecondary` later on

react/components/Modal/modal.css
thayannevls: Hey @Klynger , I just spoke with @oldufre and she confirmed about the current sizes are the ideal for the Modal. We also showed these sizes in the Styleguide Meeting and everyone agreed too

react/components/ModalDialog/index.tsx
thayannevls: I agree but I just maintained like it was before because this component will be deprecated soon and I don't pretend to add new things or change it

react/components/ModalDialog/index.tsx
thayannevls: The lint doesn't like this :rofl:

Klynger: :watt:

Klynger: Actually typescript doesn't like it, right? It shouldn't be the case, but ok then

thayannevls: Actually it is the typescript. The `onClick` of the Button doesn't accept a function that returns unknown

react/components/Modal/README.md
```diff
@@ -1,386 +1,480 @@
...
+        </div>
+      </Modal>
+    </>
+  )
+}
+
+;<ModalExample />
+```
+
+#### Sizes
```
jonathasbsouza, line 282: I think using different size buttons doesn't make clear the difference between modal sizes. I would use the same structure as in [Alert](https://styleguide.vtex.com/#/Components/Notification/Alert) scenarios, splitting each **size** alternative, and giving more context (size, usage examples) for each one.

react/components/Modal/README.md
```diff
@@ -1,386 +1,480 @@
...
+        <div className="mb3" />
+      </Modal>
+    </>
+  )
 }
-
 ;<ModalExample />
 ```
+
+## Acessibility
```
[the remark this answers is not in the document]
thayannevls: My bad! Fixed it

react/components/Modal/README.md
thayannevls: Thanks for the suggestion. I separated and improved the examples for each size, can you check?

thayannevls: https://styleguide-git-feat-new-modal-with-typescript.styleguide-core.now.sh/#/Components/Overlays/Modal

jonathasbsouza: Awesome!

react/components/Modal/index.tsx
```diff
@@ -0,0 +1,265 @@
...
+          onAnimationEnd={handleAnimationEnd}
+          data-testid="modal__overlay"
+          role="presentation"
+        >
+          <FocusTrap>{children}</FocusTrap>
+        </div>,
+        container ?? document.body
```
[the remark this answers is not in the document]
thayannevls: When `canUseDOM` is false the modal won't open because `createPortal` doesn't work :worried: it's a use case that is also happening in the current modal and I'm still not sure how to fix it

react/components/Modal/BottomBar.tsx
```diff
@@ -0,0 +1,32 @@
+import React, { forwardRef } from 'react'
+import classNames from 'classnames'
+
+export interface Props {
+  showBorder?: boolean
+  responsiveFullScreen?: boolean
+  children?: React.ReactNode
+}
+
+function BottomBar(
```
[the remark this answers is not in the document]
thayannevls: Done! I'm not using an arrow function here because will not show the component name in dev tools

react/components/Modal/TopBar.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+export interface Props {
+  showCloseIcon?: boolean
+  showTopBar: boolean
+  children: ReactNode
+  onClose: () => unknown
+  responsiveFullScreen?: boolean
+}
+
+function TopBar(
```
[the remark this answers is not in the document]
thayannevls: Done!

react/components/Modal/index.tsx
```diff
@@ -0,0 +1,265 @@
...
+
+    return () => {
+      if (
+        focusReturnNode instanceof HTMLElement &&
+        document.contains(focusReturnNode)
+      ) {
+        focusReturnNode?.focus()
+      }
+    }
+  })
```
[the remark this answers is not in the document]
thayannevls: Fixed it!

react/components/Modal/FocusTrap.tsx
```diff
@@ -0,0 +1,93 @@
...
+      handleTab(event)
+    }
+  }
+
+  useEffect(() => {
+    if (canUseDOM) document.addEventListener('keydown', handleKeyEvent)
+    return () => {
+      if (canUseDOM) document.removeEventListener('keydown', handleKeyEvent)
+    }
+  })
```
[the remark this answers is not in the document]
thayannevls: Well noticed! Thanks for the tip, just fixed it

react/components/Modal/TopBar.tsx
```diff
@@ -0,0 +1,55 @@
...
+  { showCloseIcon, children, onClose, showTopBar }: Props,
+  forwardedRef: React.Ref<HTMLDivElement>
+) {
+  const handleKeyDown = ({ key }: React.KeyboardEvent) => {
+    const SPACE = ' '
```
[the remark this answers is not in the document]
thayannevls: Done

thayannevls:
@gtkatakura Thanks for the review! Can you check it again?

thayannevls:
@matheusps I'm doing it but I will wait for @gtkatakura review because since I'll rename all the files it can be confused to track the changes requested

gtkatakura:
oh, so sorry for forgetting the review @thayannevls, you marked me but I don't come back here ... I always lost myself in the github notifications and just see now so many months ago when trying to cleaning the notifications ;(

PR #1178, opened 2020-04-28
Remove ActionMenu and Menu default fixed width

estacioneto:

As the title says, these components have fixed default widths and that's not good at all.

When all menu options have smaller widths than the default, the menu is weird.

[Running workspace](https://logistics--ambienteqa.myvtex.com/admin/shipping-strategy/warehouses/)

##### Before

image

##### After

image

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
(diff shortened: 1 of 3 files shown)


vlaux:
Probably worth it to update this (not default anymore)
https://github.com/vtex/styleguide/pull/1178/files#diff-baba83a89aadf73e148ad223e8764807R216

Also, it works smoothly if `width` prop is not defined, right?

PR #1191, opened 2020-05-06
Radio group error props

laerc:

Highlight the radio group border according to the state(whether there is an error or not).

[Running workspace](https://seller3--sandboxintegracao.myvtex.com/admin/new-seller/custom/)

The `<RadioGroup>` component does not have props that indicate errors. We are developing a product and when the user does not fill the field(RadioGroup in this case), we need to change the border color to red indicating that an error happened. To solve this problem we will add two props, `error` and `errorMessage` that indicate respectively whether there is an error or not and the error message itself.

1. Go to this [running workspace](https://seller3--sandboxintegracao.myvtex.com/admin/new-seller/custom/)
2. Then press the button `Continue`;
3. Then you should see the error displayed on the field `Type *`.

image

#### Types of changes

### DIFF
--- a/react/components/RadioGroup/index.js
+++ b/react/components/RadioGroup/index.js
@@ -24,7 +24,7 @@ class RadioGroup extends React.Component {
     } = this.props
 
     const large = size === 'large'
-    const borderColor = (error || errorMessage) ? 'b--danger hover-b--danger' : 'b--muted-4'
+    const errorHighlight = error || errorMessage
 
     return (
       <div>
@@ -47,8 +47,10 @@ class RadioGroup extends React.Component {
             const id = `${name}-${i}`
             return (
               <label
-                className={`db br3 ${borderColor} ${classNames({
+                className={`db br3 ${classNames({
                   'ba pv2 ph4': !hideBorder,
+                  'b--danger hover-b--danger': errorHighlight,
+                  'b--muted-4': !errorHighlight,
                   pointer: !isDisabled,
                 })}`}
                 key={id}
--- a/react/components/RadioGroup/index.js
+++ b/react/components/RadioGroup/index.js
@@ -47,12 +47,12 @@ class RadioGroup extends React.Component {
             const id = `${name}-${i}`
             return (
               <label
-                className={`${classNames('db br3', {
+                className={classNames('db br3', {
                   'ba pv2 ph4': !hideBorder,
                   'b--danger hover-b--danger': errorHighlight,
                   'b--muted-4': !error && !errorMessage,
                   pointer: !isDisabled,
-                })}`}
+                })}
                 key={id}
                 style={{
                   ...(!isFirst && {
--- a/react/components/RadioGroup/index.js
+++ b/react/components/RadioGroup/index.js
@@ -47,10 +47,10 @@ class RadioGroup extends React.Component {
             const id = `${name}-${i}`
             return (
               <label
-                className={`db br3 ${classNames({
+                className={`${classNames('db br3', {
                   'ba pv2 ph4': !hideBorder,
                   'b--danger hover-b--danger': errorHighlight,
-                  'b--muted-4': !errorHighlight,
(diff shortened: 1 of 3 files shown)


react/components/RadioGroup/index.js
```diff
@@ -19,9 +19,12 @@ class RadioGroup extends React.Component {
       label,
       testId,
       size,
+      error,
+      errorMessage,
     } = this.props
 
     const large = size === 'large'
+    const borderColor = (error || errorMessage) ? 'b--danger hover-b--danger' : 'b--muted-4'
```
tkovs, line 27: I had doubts about this condition, but our `Input` works in the [same way](https://github.com/vtex/styleguide/blob/f67a7e2967db7f16b2476ee68c2871712b699f2b/react/components/Input/index.js#L138).

react/components/RadioGroup/index.js
```diff
@@ -44,8 +47,8 @@ class RadioGroup extends React.Component {
             const id = `${name}-${i}`
             return (
               <label
-                className={`db br3 ${classNames({
-                  'b--muted-4 ba pv2 ph4': !hideBorder,
+                className={`db br3 ${borderColor} ${classNames({
+                  'ba pv2 ph4': !hideBorder,
                   pointer: !isDisabled,
                 })}`}
```
estacioneto, line 53:
```suggestion
                className={classNames('db br3', {
                  'b--danger hover-b--danger': error || errorMessage,
                  'b--muted-4': !error && !errorMessage,
                  'ba pv2 ph4': !hideBorder,
                  pointer: !isDisabled,
                })}
```

Also there could be a variable `const errorHighlight = error || errorMessage` to make it cleaner

react/components/RadioGroup/index.js
```diff
@@ -75,6 +78,9 @@ class RadioGroup extends React.Component {
               </label>
             )
           })}
+          {errorMessage && (
+            <div className="c-danger t-small mt3 lh-title">{errorMessage}</div>
```
estacioneto, line 82: Also, shouldn't it be a `span` element?

react/components/RadioGroup/index.js
laerc: Yeah, me too. I have used the `Input` as an example to keep consistency.

react/components/RadioGroup/index.js
laerc: I have used `Input` as an example to keep consistency. I think the reason for using `<div>` here is to render the error by block to not render inline with other possible components bellow.

react/components/RadioGroup/index.js
laerc: Nice. I have made the changes ✨

laerc:
> Nice! Just missing an example in [README.md](https://github.com/vtex/styleguide/blob/master/react/components/RadioGroup/README.md).

Added the example 🚀

react/components/RadioGroup/index.js
```diff
@@ -44,8 +47,10 @@ class RadioGroup extends React.Component {
...
             return (
               <label
-                className={`db br3 ${classNames({
-                  'b--muted-4 ba pv2 ph4': !hideBorder,
+                className={`${classNames('db br3', {
+                  'ba pv2 ph4': !hideBorder,
+                  'b--danger hover-b--danger': errorHighlight,
+                  'b--muted-4': !error && !errorMessage,
                   pointer: !isDisabled,
                 })}`}
```
estacioneto, line 55: 😅 
```suggestion
                className={classNames('db br3', {
                  'ba pv2 ph4': !hideBorder,
                  'b--danger hover-b--danger': errorHighlight,
                  'b--muted-4': !error && !errorMessage,
                  pointer: !isDisabled,
                })}
```

laerc: I had tried without this and kind of broke the styles. But I have tested now and it works perfectly, that's strange hahaha but I will change to this.

PR #1192, opened 2020-05-07
Update Alert docs #trivial

jonathasbsouza:

Update `Alert` examples.

There are mixed `alert` examples without e-commerce context. Also, we don't properly explain each component variant usage.

`yarn styleguide`

#### Types of changes


react/components/Alert/README.md
```diff
@@ -14,47 +14,59 @@
 
 For low-priority notification prefer a <a href="#/Components/Notification/ToastProvider">Toast</a>.
 
+## Examples
```
kevinch, line 17: We do not use such title in the other components docs. I'd use a space or maybe `hr` to separate examples from explanations above.

jonathasbsouza: Got it! Changed it in deb0abc

kevinch: Nice!

PR #1194, opened 2020-05-08
Add useDisclosure hook

thayannevls:
As the title says.

Some components deal with the scenario of open, close, and toggle behavior.

`yarn && yarn styleguide`
#### Types of changes

### DIFF
--- a/react/utilities/index.ts
+++ b/react/utilities/index.ts
@@ -1,0 +1,12 @@
+import { useState } from 'react'
+
+export const useDisclosure = (open = false) => {
+  const [isOpen, setIsOpen] = useState(open)
+
+  return {
+    isOpen,
+    onOpen: () => setIsOpen(true),
+    onClose: () => setIsOpen(false),
+    onToggle: () => setIsOpen(!isOpen),
+  }
+}
--- a/react/utilities/index.ts
+++ b/react/utilities/index.ts
@@ -1,12 +1,16 @@
-import { useState } from 'react'
+import { useState, useCallback } from 'react'
 
 export const useDisclosure = (open = false) => {
   const [isOpen, setIsOpen] = useState(open)
 
+  const onOpen = useCallback(() => setIsOpen(true), [])
+  const onClose = useCallback(() => setIsOpen(false), [])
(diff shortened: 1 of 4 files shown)


react/utilities/index.ts
```diff
@@ -0,0 +1,12 @@
+import { useState } from 'react'
+
+export const useDisclosure = (open = false) => {
+  const [isOpen, setIsOpen] = useState(open)
+
+  return {
+    isOpen,
+    onOpen: () => setIsOpen(true),
+    onClose: () => setIsOpen(false),
+    onToggle: () => setIsOpen(!isOpen),
```
matheusps, line 10:
```suggestion
    onToggle: () => setIsOpen(prevIsOpen => !prevIsOpen),
```

matheusps, line 10: Wrap these in `useCallback`. This is useful if the users may want to add then to dependency arrays.

styleguide.config.js
```diff
@@ -214,6 +214,15 @@ module.exports = {
         },
       ],
     },
+    {
+      name: 'Utilities',
```
matheusps, line 218: Hooks maybe?

react/utilities/index.ts
thayannevls: Done! :smile:

thayannevls: Done!

styleguide.config.js
thayannevls: :thinking: Makes sense

@emersonlaurentino do you agree?

react/utilities/index.ts
```diff
@@ -0,0 +1,12 @@
+import { useState, useCallback } from 'react'
+
+export const useDisclosure = (open = false) => {
+  const [isOpen, setIsOpen] = useState(open)
+
+  return {
+    isOpen,
+    onOpen: useCallback(() => setIsOpen(true), []),
```
matheusps, line 8: Doesn't this usage break the first rule of hooks? 🤔.

thayannevls: Fixed it

styleguide.config.js
emersonlaurentino: I believe not, because not all utilities are hooks

PR #1196, opened 2020-05-12
Render empty span if atom has no value

rerissondaniel:

As the title says.
[Running workspace](https://emptyatom--storecomponents.myvtex.com/admin/collections/1938/)
Before this change:

Captura de tela de 2020-05-12 12-45-43
Captura de tela de 2020-05-12 12-46-18

After this change:
Captura de tela de 2020-05-12 12-49-20
Captura de tela de 2020-05-12 12-49-48

#### Types of changes

### DIFF
--- a/react/components/FilterBar/FilterTag.js
+++ b/react/components/FilterBar/FilterTag.js
@@ -144,6 +144,7 @@ class FilterTag extends PureComponent {
       newFilterLabel,
       isMobile,
       testIds,
+      noOptionsMessage,
     } = this.props
     const { isMenuOpen, virtualStatement } = this.state
 
@@ -271,6 +272,7 @@ class FilterTag extends PureComponent {
                   omitVerbs={shouldOmitVerb}
                   options={compatibleOptions}
                   subjectPlaceholder={subjectPlaceholder}
+                  noOptionsMessage={noOptionsMessage}
                   statement={
                     isMoreOptions
                       ? virtualStatement
@@ -339,6 +341,7 @@ FilterTag.propTypes = {
   newFilterLabel: PropTypes.string,
   device: PropTypes.string,
   isMobile: PropTypes.bool,
+  noOptionsMessage: PropTypes.func,
   testIds: PropTypes.shape({
     moreOptionsButton: PropTypes.string,
     submitFiltersButton: PropTypes.string,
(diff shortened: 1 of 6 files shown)


react/components/Statement/Atoms/VerbAtom.tsx
```diff
@@ -44,6 +44,8 @@ const VerbAtom: React.FC<Props> = ({
         <span className="pt3">{value.label}</span>
       )}
     </div>
+  ) : (
+    <span className="pt3" />
```
kevinch, line 48: Why not null or Fragment?

rerissondaniel: Without the span to apply this margin it gets little bit weird:

Captura de tela de 2020-05-14 11-13-50

I could fix that margin in the specific app I'm using it but probably would require to fix it in other apps that are using it too

react/components/FilterBar/FilterTag.js
```diff
@@ -339,6 +341,7 @@ FilterTag.propTypes = {
   newFilterLabel: PropTypes.string,
   device: PropTypes.string,
   isMobile: PropTypes.bool,
+  noOptionsMessage: PropTypes.func,
```
estacioneto, line 344: Why a function? How would that be used?

react/components/Statement/Atoms/VerbAtom.tsx
estacioneto: Why not solve this in `Statement` component?

rerissondaniel: Because the value I need to display is calculated in this component and theres a lot of logic in the statement already

react/components/FilterBar/FilterTag.js
rerissondaniel: It receives the value and returns the message to be displayed. This is just forwarded to the library used for the select

react/components/Statement/Atoms/VerbAtom.tsx
estacioneto: The amount of logic in the `Statement` component is, in my opinion, irrelevant for this particular case. About the value calculated I agree that it looks like the logic should be inside but it doesn't. The `verbs.length > 0 && (<Verb />)` condition should solve the problem

rerissondaniel: Ok, I did like that but theres the problem with the margin. I'll fix it in admin collections but I don't know what impact this will have in other apps

estacioneto: I think that to solve this margin problem you could extract the element inside the `VerbAtom` to the `Statement` component. This way you could let the container with this margin and conditionally show the `VerbAtom`. What do you think?

PR #1203, opened 2020-05-18
Make alert accessible

felippenardi:
This is a non-breaking change to add some accessibility features to the alert.

[Running workspace](https://accessibility--content.myvtex.com/admin/app/new-cms/edit/new)

1- Allow developers to use `focusOnOpen` property to make the cursor focus change to either the action or dismiss buttons
2- Make screen readers announce the alert error as soon as it pops up in the screen

You can go to the workspace link above and try saving a document without filling any information. This will trigger an error. Then you should try repeating the action and, in the second time, dismissing the alert.

You should be able to do all this using just your keyboard.

If you have voice over activated, you should be able to hear the entire experience seamlessly.

alert

#### Types of changes

### DIFF
--- a/react/components/Alert/index.js
+++ b/react/components/Alert/index.js
@@ -9,18 +9,37 @@ import Button from '../Button'
 import { withForwardedRef, refShape } from '../../modules/withForwardedRef'
 
 class Alert extends Component {
+  constructor(props) {
+    super(props)
+    this.firstElementRef = React.createRef()
+  }
+
   componentDidMount() {
     if (this.props.autoClose && this.props.onClose) {
       this.timeout = setTimeout(this.props.onClose, this.props.autoClose)
+    }
+
+    if (this.props.focusOnOpen) {
+      this.firstElementRef.current && this.firstElementRef.current.focus()
     }
   }
 
   componentWillUnmount() {
     clearTimeout(this.timeout)
+
+    if (this.props.focusAfterClosed && this.props.focusAfterClosed.current) {
+      this.props.focusAfterClosed.current.focus()
+    }
   }
 
   render() {
-    const { type, onClose, action, forwardedRef } = this.props
+    const {
+      type,
+      onClose,
+      action,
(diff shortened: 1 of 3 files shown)


felippenardi:
I'll add the documentation and changelog in a bit

emersonlaurentino:
@felippenardi great job, we need more a11y components. 

Can you add changelog?

react/components/Alert/index.js
```diff
@@ -109,11 +137,17 @@ Alert.propTypes = {
   onClose: PropTypes.func,
   /** Time in ms to auto close the alert */
   autoClose: PropTypes.number,
+  /** Set focus to the first focusable element inside alert, which should be the "action" when available or the "close" button */
+  focusOnOpen: PropTypes.bool,
+  /** Define element to be focused when the alert is dimissed */
+  focusAfterClosed: PropTypes.shape({ current: PropTypes.instanceOf(Element) }),
```
Klynger, line 143: I think this prop is not needed, you can get `document.activeElement` and hold so when the user dismisses the alert its focuses on this element. Another thing to think about is when the user dismiss the alert, but the focus wasn't in the alert anymore, so you shouldn't mess with the current focused element

felippenardi: Just talked to Klynger on discord, and he suggested the idea of replacing both these properties for a single property that receives an ENUM.

The enum would have three options:
- Focus on action or dismiss when the alert opens
- Do the above and also register the document.activeElement and return to it when dismiss is triggered

felippenardi: After analyzing the behavior I noticed a few scenarios where changing the focus after closing may be problematic, so I've decided not to implement it.

felippenardi:
@emersonlaurentino @Klynger @iaronaraujo I've removed the focus feature for when the alert is dismissed because it would potentially cause accessibility problems.

I've also added changelog, let me know if we are good to merge.

PR #1206, opened 2020-05-19
Add TrapFocus component in Modal and fix Input autoFocus

thayannevls:
Implement our own solution to trap focus in Modal.

Dialogs must have focus trapped inside.

`yarn && yarn storybook`

1. Go to http://localhost:6006/?path=/story/components-modal--default
2. Navigate through Modal with `TAB` and `SHIFT+TAB` and check if it is trapped inside modal.

#### Types of changes

### DIFF
--- a/react/components/Modal/FocusTrap.tsx
+++ b/react/components/Modal/FocusTrap.tsx
@@ -31,10 +31,9 @@ const focusFirstElement = (element: HTMLElement) => {
 
 interface Props {
   children?: React.ReactNode
-  initialFocusRef?: React.RefObject<HTMLElement> | null
 }
 
-const FocusTrap: FC<Props> = ({ children, initialFocusRef }) => {
+const FocusTrap: FC<Props> = ({ children }) => {
   const focusContainer = useRef<HTMLDivElement>(null)
 
   const handleFocusIn = (event: FocusEvent) => {
@@ -92,12 +91,11 @@ const FocusTrap: FC<Props> = ({ children, initialFocusRef }) => {
     if (!focusContainer.current) {
       return
     }
-    if (initialFocusRef) {
-      initialFocusRef?.current?.focus()
-      return
-    }
-    focusFirstElement(focusContainer.current)
-  }, [focusContainer, initialFocusRef])
+    const alreadyHasFocus =
+      focusContainer.current &&
+      focusContainer.current.contains(document.activeElement)
+    if (!alreadyHasFocus) focusFirstElement(focusContainer.current)
+  }, [focusContainer])
 
   return (
     <>
--- a/react/components/Modal/FocusTrap.tsx
+++ b/react/components/Modal/FocusTrap.tsx
@@ -1,0 +1,100 @@
+import React, { FC, useEffect, useRef } from 'react'
+
+enum Key {
+  TAB = 9,
+  SHIFT = 16,
+}
+
+const FOCUSABLE_SELECTOR =
+  'a,frame,iframe,input:not([type=hidden]):not(:disabled),select:not(:disabled),textarea:not(:disabled),button:not(:disabled),*[tabindex]:not([tabindex="-1"])'
+
+const findFirstFocusable = (element: HTMLElement): HTMLElement | null => {
+  if (element.matches && element.matches(FOCUSABLE_SELECTOR)) {
+    return element
+  }
+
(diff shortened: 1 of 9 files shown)


react/components/Modal/FocusTrap.tsx
```diff
@@ -0,0 +1,109 @@
...
+      initialFocusRef?.current?.focus()
+      return
+    }
+    focusFirstElement(focusContainer.current)
+  }, [focusContainer, initialFocusRef])
+
+  return (
+    <div ref={focusContainer} {...props}>
+      {children}
+    </div>
```
Klynger, line 105: Instead of adding this aditional div you could just get the ref from children and work with taht

react/components/Modal/README.md
```diff
@@ -32,11 +32,14 @@
 | aria-label              | `string`  | 🚫       |                     | Acessible Modal name. If this name is visible on the screen, prefer to use aria-labelledby                                                                  |
 | aria-labelledby         | `string`  | 🚫       | `vtex-modal__title` | ID of the element that provides the Modal an accessible name. If aria-label and aria-albelledby is not defined, the default here will be the title element. |
 | aria-describedby        | `string`  | 🚫       |                     | ID of the element that provides the Modal an accessible description                                                                                         |
+| initialFocusRef         | `ref`     | 🚫       |                     | Element to be focused first when Modal is open. The default is the first Modal element.                                                                     |
```
Klynger, line 35: I think the user can do this without having to tell the modal to do it

react/components/Modal/FocusTrap.tsx
thayannevls: How can I do this? With `cloneElement`?

Klynger: Yes, `cloneElement(children, { ref: childRef })

react/components/Modal/README.md
thayannevls: Yes, it can. But it's not that simple because of the Modal behavior of opening, the user needs to do a [workaround](https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node) to check if `ref.currrent` changes since `useEffect(effect, [ref.current])` doesn't work.

This prop is an easy way for the user to control the initial focus, [ReachUI](https://reacttraining.com/reach-ui/dialog/#dialogoverlay-initialfocusref) does this too.

Klynger: You have a test to do this, I got your test and made how the user could do it

```typescript
  describe('initialFocusRef', () => {
    it.only('to focus correctly', () => {
      const containerModal = document.createElement('div')
      const onClose = jest.fn()

      const Component = () => {
        const initialFocusRef = useRef<HTMLDivElement>(null)

        useEffect(() => {
          initialFocusRef.current?.focus()
        }, [])

        return (
          <Modal
            isOpen
            onClose={onClose}
            container={containerModal}
          >
            <Input
              placeholder="Type your name..."
              size="small"
              ref={initialFocusRef}
            />
          </Modal>
        )
      }
      render(<Component />, { container: containerModal })

      const currentFocused = document.activeElement

      expect(currentFocused).toBe(document.querySelector('input'))
    })
  })
```

Klynger: My problem with this prop is that it is a bit strange to tell the modal what you want o focus and another thing is what would happen if I pass the prop and did the `userEffect` that I added to the test for a different element? Who would be focused? Is a bit confusing the usage of it

thayannevls: I think it works because the modal is already open :thinking: I did the same thing at the storybook and didn't work

thayannevls: @Klynger Fixed and removed the prop. But I had to do a small fix on the Input component to work properly

react/components/Modal/modal.stories.tsx
```diff
@@ -64,7 +66,12 @@ export const Default = () => {
         This is a simple customizable Modal to ask your name! Please try change
         my props below in the Knobs and see how I react!
         <div className="mt6 mb5">
-          <Input placeholder="Type your name..." size="small" />
+          <Input
+            // eslint-disable-next-line jsx-a11y/no-autofocus
```
matheusps, line 70: This should be handled by focusing the ref on `useLayoutEffect`

PR #1221, opened 2020-05-26
Fix AutocompleteInput disabled style

thayannevls:
Add a disabled background when the `AutocompleteInput` is disabled

`yarn && yarn styleguide`
Change disabled to `true` in an AutocompleteInput example.

**Before**
Screenshot from 2020-05-26 16-07-35

**After**
Screenshot from 2020-05-26 16-24-04

#### Types of changes


jonathasbsouza:
@thayannevls what do you think of creating a section in Autocomplete docs showing that it has the same states of Input?

thayannevls:
@jonathasbsouza It would be great, I'm not sure how would be the session but we can work together to do this. But I prefer to do this in a separate PR if you agree to too

PR #1240, opened 2020-06-04
Do not break if ref is not present

klzns:

Fix issue 
image

[Running workspace](http://link)

#### Types of changes


react/components/Checkbox/index.js
```diff
@@ -101,7 +101,7 @@ class Checkbox extends PureComponent {
               elem && (this.myCheckbox = elem)
               if (typeof forwardedRef === 'function') {
                 forwardedRef(elem)
-              } else if (typeof forwardedRef === 'object') {
+              } else if (forwardedRef) {
```
Klynger, line 104:
```suggestion
              } else if (forwardedRef && typeof forwardedRef === 'object') {
```

PR #1247, opened 2020-06-12
Fix FilterTag disabled style

thayannevls:
Fix FilterTag disabled style when there is a statement selected.

`IconClose` being active even through the filter tag is disabled

`yarn && yarn styleguide`

**Before**
Screenshot from 2020-06-12 16-58-44

**After**
Screenshot from 2020-06-12 16-58-05

#### Types of changes

### DIFF
--- a/react/components/FilterBar/FilterTag.js
+++ b/react/components/FilterBar/FilterTag.js
@@ -203,12 +203,14 @@ class FilterTag extends PureComponent {
           pr4: !isEmpty && !isMoreOptions,
           'hover-bg-muted-5 b--muted-4':
             (alwaysVisible && isEmpty) || isMoreOptions,
-          'bg-action-secondary hover-bg-action-secondary b--action-secondary': !(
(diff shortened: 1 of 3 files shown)


react/components/FilterBar/FilterTag.js
```diff
@@ -315,8 +317,14 @@ class FilterTag extends PureComponent {
...
           {!isEmpty && !isMoreOptions && (
             <div
-              className="flex items-center c-link hover-c-link pointer"
+              className={classNames('flex items-center', {
+                'c-link hover-c-link pointer': !disabled,
+                'bg-disabled': disabled,
+              })}
               onClick={() => {
+                if (disabled) {
+                  return
```
kevinch, line 326: That means that the user clicks but nothing happens right? Any given feedback?

emersonlaurentino: This is the standard behavior of a button, but as we are using a div instead of a button, we need to do this.

PR #1249, opened 2020-06-16
Adjust react-select style

vlaux:
Add some custom styles to react-select in order to correspond to our design system style.

Since height, padding and font size from select were different from other inputs, they use to get unaligned when together:
image
Placeholder font size:
image

[Running workspace](https://vlauxbeta--pricingqa.myvtex.com/admin/promotions/1a18dd71-26ae-428b-a425-af6d816af784/)

image

image

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `FilterBar` clear icon disabled style.
 - `AutocompleteInput` disabled style.
 - Filters hidden for existing statements.
+- `EXPERIMENTAL_Select` style.
 
 ### Added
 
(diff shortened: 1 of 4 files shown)


vlaux:
There may be an issue with font size for different select's `size` variants.
Since `regular` is the default, should we only apply the input's default font size to this version, while `small` and `large` use other font sizes? cc @jonathasbsouza

vlaux:
After talking to @jonathasbsouza, I decided to convert this to draft until we fix all size variants (small, regular, large) to correspond our updated input's style

davicosta99:
We discussed the component at the working session. Here's a summary:
- Tag vertical alignment. 16pt font-size in the large variant. (Maybe reuse the tag component)
- Focus state and error state not aligned with the Input component. Focus state shouldn't have a hover behavior.
- Ideally, when a tag was at the beginning, the field's left padding would be 8px, instead of 16px.

vlaux:
So, I did some adjustments:
* Fixed hover & focus border color
* Fixed error state
* Using 16px font size on tags for `large` variant
* Reduced left padding on multi input selects from 16px to 4px (as discussed with @davicosta99, 8px was not enough)

I couldn't really solve the vertical alignment in a consistent way, so I let it as it is for now. If someone wants to take a look on it, feel free to add it!

Issue #1244, opened 2020-06-10
Use size prop on AutocompleteInput


edenalsant:
**Describe the bug**
In the [AutocompleteInput](https://styleguide.vtex.com/#/Components/Forms/AutocompleteInput) component there's no size related prop, making harder to adjust it's size. 

**To Reproduce**
**Steps to reproduce the behavior:**
With the [InputSearch](https://styleguide.vtex.com/#/Components/Forms/InputSearch) component it's possible to adjust it's size as it has a size props exposed. 
To see the different among them
Go to this  [page](https://help.vtex.com/). Here it's used the `InputSearch` component, whereas in [this workspace](https://autocomplete--vtexhelp.myvtex.com/) it's being used the `AutcompleteInput`.

**Expected behavior**
 The adjustment of the component's size being done by setting the prop

**Screenshots**
Production environment using `InputSearch`
![image](https://user-images.githubusercontent.com/29438497/84289654-22869980-ab19-11ea-8d82-435a2996e139.png)

Workspace using `AutocompleteInput` 
![image](https://user-images.githubusercontent.com/29438497/84289702-316d4c00-ab19-11ea-835d-d3890eae5e21.png)

vlaux:
Could this issue be closed due to #1251 already merged into master? cc @edenalsant

PR #1251, opened 2020-06-17
Feature/add size prop

edenalsant:

Add display prop to make AutocompleteInput size be determined by its value

[issue 1244](https://github.com/vtex/styleguide/issues/1244)
The static definition of the [AutocompleteInput component](https://styleguide.vtex.com/#/Components/Forms/AutocompleteInput)'s size

Clone this repo and run
`yarn && yarn styleguide`
go to `localhost:6060`

In the file `react/components/AutcompleteInput/README.md` you can set the prop `display` to either value of `small`, `regular`, `large` or you can simply ignore this prop (default is set to `regular`). 
You should see the bar changing according to the specified size

image
small - regular - large (respectively) 

#### Types of changes


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- 'display' prop to AutocompleteInput
```
ullayne02, line 12: what do think about the name of the props `size`?

edenalsant: just did it! the prop name exposed to `AutocompleteInput` is now `size`. However, I was't able to set the same name for the `SearchInput` component. My guess is it's somehow already used, so I tried to use a more descriptive name (`barSize`, btw)there even though they're not the same

react/components/AutocompleteInput/SearchInput/index.tsx
```diff
@@ -24,10 +24,13 @@ const propTypes = {
   onBlur: PropTypes.func,
   /** Determine if the input and the button should be disabled */
   disabled: PropTypes.bool,
+  /** Determine the search bar size */
+  barSize: PropTypes.string,
```
analuizamtg, line 28: shouldn't this be `size`?

edenalsant: I explained why I used `barSize` instead [here](https://github.com/vtex/styleguide/pull/1251#discussion_r441763471)

estacioneto: This will do it
```ts
const SearchInput: React.FC<PropTypes.InferProps<typeof propTypes> &
  Omit<
    React.HTMLProps<HTMLInputElement>,
    'onChange' | 'value' | 'size'
  >> = props => {
// ....
```

edenalsant: it worked! :awthanks:

react/components/AutocompleteInput/SearchInput/index.tsx
```diff
@@ -24,14 +24,20 @@ const propTypes = {
   onBlur: PropTypes.func,
   /** Determine if the input and the button should be disabled */
   disabled: PropTypes.bool,
+  /** Determine the search bar size */
+  size: PropTypes.string,
```
estacioneto, line 28: How about...

```suggestion
  size: PropTypes.oneOf(['small', 'regular', 'large']),
```

react/components/AutocompleteInput/index.tsx
```diff
@@ -71,6 +71,11 @@ const propTypes = {
       /** Last Searched options's title */
       label: PropTypes.node.isRequired,
     }),
+    /**
+     * Selects a size of the input bar, could be set to `small`, `regular` or `large`.
+     * `regular` is the default value.
+     */
+    size: PropTypes.string,
```
estacioneto, line 78: Same here
```suggestion
    size: PropTypes.oneOf(['small', 'regular', 'large']),
```

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- 'size' prop to AutocompleteInput
```
thayannevls, line 12: Just a suggestion:
```suggestion
- `size` prop to AutocompleteInput
```

react/components/AutocompleteInput/SearchInput/index.tsx
thayannevls: I prefer the name `medium` instead of `regular` because it's more clarifier :thinking: what exactly is regular?

react/components/AutocompleteInput/SearchInput/index.tsx
```diff
@@ -74,7 +81,7 @@ const SearchInput: React.FC<PropTypes.InferProps<typeof propTypes> &
 
   const buttonClasses = classNames(
     activeClass,
-    'bg-base br2 br--right h-regular w3 bw1 ba pa0 bl-0',
+    `bg-base br2 br--right w3 bw1 ba pa0 bl-0' h-${size}`,
```
thayannevls, line 84: You need to check if the size is valid to prevent a broken style, if it's not one of `small`, `regular` or `large` change it to `regular`

react/components/AutocompleteInput/README.md
```diff
@@ -39,6 +39,7 @@ const UsersAutocomplete = () => {
             ? user.toLowerCase().includes(term.toLowerCase())
             : user.label.toLowerCase().includes(term.toLowerCase())
         ),
+    size: 'small',
```
thayannevls, line 42: Can you add an isolate example of sizes to this component?

Examples:
Screenshot from 2020-06-18 17-49-27
Screenshot from 2020-06-18 17-49-54

react/components/AutocompleteInput/SearchInput/index.tsx
matheusps: Just `size='regular'` as default prop is enough, IMO. :)

react/components/AutocompleteInput/SearchInput/index.tsx
edenalsant: `regular` seems to be used all across the components to define the "medium" size 🤔  Using `medium` instead would differ from the rest of them

react/components/AutocompleteInput/SearchInput/index.tsx
thayannevls: But what if the component receives a size that doesn't exist? Example `size="aa"`

edenalsant: when we a have a non-existent size passed to the prop, we get this behaviour: 
image

the bar gets smaller that the small sized bar, and we also get a warn about the invalid prop:

image

From the development perspective, isn't better for the developer to see that something might be wrong on the console instead of having that wrong value being replaced by the `regular` size? I'm thinking of something like, If he/she passed a typo, for example, we would replace it for regular and it would be warned anywhere as we're fixing it behind the scenes.

@thayannevls @matheusps

ullayne02: I agree with Eden

thayannevls: I agree that must have the warning in the console, but not to let the component have a random size like it is now :thinking: other components in styleguide when receiving an invalid size, it forces to go to the `regular` and throws a warning in the console.

matheusps: I agree @edenalsant

thayannevls: My only concern is that the component is getting smaller than the `small` size when something is wrong

react/components/AutocompleteInput/README.md
```diff
@@ -252,3 +252,177 @@ const UsersAutocomplete = () => {
 
 ;<UsersAutocomplete />
 ```
+
+#### Size
+
+### small size bar
```
matheusps, line 258: I would put the three inputs in the same example. You don't need to repeat all that logic, IMO.

react/components/AutocompleteInput/SearchInput/index.tsx
edenalsant: but wouldn't it be a indication that something is wrong?

thayannevls: I don't think that getting the component to have a random size is a good indication when something is wrong, you have the error in the console for this. You can check what other components with sizes in styleguide to compare

edenalsant: done ✨  @thayannevls

matheusps: @edenalsant. I've discussed with @thayannevls thought a little bit more about this, and she's right. The ideal way to handle this is not just to fill the `height-x` className, but a function that does not allow you to have `height-undefined`. But, you're maybe wondering "it is never undefined" 🤔. In the way you're doing right now, kinda is! The `height-wrongWord` is not defined by the design language, so you render an input with a wrong look (I didn't notice that the inputs had different heights on your print, sorry about that).

edenalsant: awesome! thank you, guys! you rock! ✨

matheusps:
You must update the test snapshots.

react/components/AutocompleteInput/README.md
```diff
@@ -252,3 +252,97 @@ const UsersAutocomplete = () => {
...
+      } else {
+        setTerm(term)
+      }
+    },
+    onSearch: (...args) => console.log('onSearch:', ...args),
+    onClear: () => setTerm(''),
+    placeholder: 'Search user... (e.g.: Ana)',
+    value: term,
+  }
+  return (<div>
```
estacioneto, line 335: A prettier would be nice...

react/components/AutocompleteInput/SearchInput/index.tsx
```diff
@@ -64,17 +70,28 @@ const SearchInput: React.FC<PropTypes.InferProps<typeof propTypes> &
...
+    case 'small':
+      activeClass += ` h-${size}`
+      break
+    case 'large':
+      activeClass += ` h-${size}`
+      break
+    default:
+      activeClass += ' h-regular'
+  }
+
```
estacioneto, line 91: How about...

```suggestion
  const regularSize = size !== 'small' && size !== 'large'
  const activeClass = classNames({
    'b--muted-3': focused,
    'b--muted-4': !focused,
    'br--top': !roundedBottom,
    'bg-disabled': disabled,
    'bg-base': !disabled,
    [`h-${size}`]: !regularSize,
    'h-regular': regularSize,
  })
  
```

react/components/AutocompleteInput/README.md
edenalsant: thank you so much for noticing

PR #1252, opened 2020-06-18
Fix the device detection in the `withDeviceHoc` that it not consistent when the application is SSR.

lucasaarcoverde:
Depends on https://github.com/vtex/admin-collections/pull/360
Fix the device detection in the `withDeviceHoc` that it not consistent when the application is SSR.

[Running workspace](https://filtertags--cosmetics1.myvtex.com/admin/app/collections/7031?tab=add)
The `withDeviceHoc` HOC that detects the device is not consistent when the application is SSR, causing re-rendering with the wrong device type as prop. To resolve this, was added a option to pass this check as a prop in the`FilterBar` component.

In the running workspace, switch from tab `Adicionar Produtos` to `Minha Coleção`
You can check an example of usage [here](https://github.com/vtex/admin-collections/blob/ca7ed3dec5bf7a620a9e62b813eb005600d3aeb6/react/components/CollectionView/ProductsContent/ProductsTable/OrganizerTable/components/FilterBar.tsx#L37-L52)
##### Before
input2
##### After
input1

##### Passing static props
- device: `phone` isMobile: `true`
Captura de tela de 2020-06-18 12-30-43
- device: `desktop` isMobile: `false`
Captura de tela de 2020-06-18 12-31-36

#### Types of changes

### DIFF
--- a/react/components/utils/withDeviceHoc.tsx
+++ b/react/components/utils/withDeviceHoc.tsx
@@ -8,18 +8,23 @@ export enum Device {
 }
 
 const withDevice = Component => {
-  function WithDevice(props) {
+  function WithDevice({
+    device: defaultIsDevice,
+    isMobile: defaultIsMobile,
+    ...props
+  }) {
     // Taken at https://github.com/vtex-apps/device-detector/blob/master/react/useDevice.tsx#L22
     const isScreenMedium = useMedia({ minWidth: '40rem' })
     const isScreenLarge = useMedia({ minWidth: '64.1rem' })
 
-    const device = isScreenLarge
-      ? Device.desktop
-      : isScreenMedium
-      ? Device.tablet
-      : Device.phone
+    const device =
+      defaultIsDevice ?? isScreenLarge
+        ? Device.desktop
+        : isScreenMedium
+        ? Device.tablet
+        : Device.phone
 
-    const isMobile = device.toLowerCase() === 'phone'
+    const isMobile = defaultIsMobile ?? device.toLowerCase() === 'phone'
 
     return <Component device={device} isMobile={isMobile} {...props} />
   }
--- a/react/components/utils/withDeviceHoc.tsx
+++ b/react/components/utils/withDeviceHoc.tsx
@@ -9,8 +9,8 @@ export enum Device {
 
 const withDevice = Component => {
   function WithDevice({
-    device: defaultIsDevice,
-    isMobile: defaultIsMobile,
+    device: customDevice,
+    isMobile: customIsMobile,
     ...props
   }) {
     // Taken at https://github.com/vtex-apps/device-detector/blob/master/react/useDevice.tsx#L22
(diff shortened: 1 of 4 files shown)


react/components/utils/withDeviceHoc.tsx
```diff
@@ -8,18 +8,23 @@ export enum Device {
 }
 
 const withDevice = Component => {
-  function WithDevice(props) {
+  function WithDevice({
+    device: defaultIsDevice,
```
estacioneto, line 12:
```suggestion
    device: defaultDevice,
```

react/components/utils/withDeviceHoc.tsx
```diff
@@ -8,18 +8,23 @@ export enum Device {
...
     const isScreenMedium = useMedia({ minWidth: '40rem' })
     const isScreenLarge = useMedia({ minWidth: '64.1rem' })
 
-    const device = isScreenLarge
-      ? Device.desktop
-      : isScreenMedium
-      ? Device.tablet
-      : Device.phone
+    const device =
+      defaultIsDevice ?? isScreenLarge
```
estacioneto, line 21: If it's supposed to be a default value, shouldn't it be on the right side of the operation?
Also, why is a string being evaluated as a boolean value?

react/components/utils/withDeviceHoc.tsx
```diff
@@ -8,18 +8,23 @@ export enum Device {
...
+        ? Device.desktop
+        : isScreenMedium
+        ? Device.tablet
+        : Device.phone
 
-    const isMobile = device.toLowerCase() === 'phone'
+    const isMobile = defaultIsMobile ?? device.toLowerCase() === 'phone'
```
estacioneto, line 27: Same here... It's a default value or `device.toLowerCase() === 'phone'` is the fallback?

react/components/utils/withDeviceHoc.tsx
lucasaarcoverde: Oh, didn't notice that. Thank you, i'll solve

react/components/utils/withDeviceHoc.tsx
lucasaarcoverde: I think that i used the wrong word to name this prop, is more to `customDevice` and `customIsMobile`, since it is passed through the `FilterBar` component

react/components/utils/withDeviceHoc.tsx
lucasaarcoverde: If the `isMobile` prop has not been set, the standard way to check the device type is using `device.toLowerCase() === 'phone'`

react/components/utils/withDeviceHoc.tsx
lucasaarcoverde: It isn't evaluated as a boolean, it checks if there is a defined `customDevice`, otherwise, it will return some type of the `Device` enum

estacioneto: Now I get it. Much better with the `custom*` thing.

PR #1262, opened 2020-06-26
Add custom option component to react-select

vlaux:
As discussed on the last Styleguide's working session, following the work done on https://github.com/vtex/styleguide/pull/1249, we would like to do some more changes on our select to make it more consistent with our Design System. 
One defined actionable was to change de dropdown behaviour to show the selected option with a checkbox in front instead of highlighting it with a primary color (`react-select`'s default).

Also fixes multi-select's tag hover ;-) 

https://styleguide-git-feature-custom-option-comp.styleguide-core.vercel.app/

Before:
image
After:
image

#### Types of changes

### DIFF
--- a/react/components/EXPERIMENTAL_Select/Option.tsx
+++ b/react/components/EXPERIMENTAL_Select/Option.tsx
@@ -4,13 +4,22 @@ import OptionProps from 'react-select'
 import IconCheck from '../icon/Check'
 
 const Option = (props: OptionProps) => {
-  const { children, getStyles, isSelected, innerRef, innerProps } = props
+  const {
+    children,
+    getStyles,
+    isMulti,
+    innerProps,
+    innerRef,
+    isSelected,
+  } = props
   return (
     <div style={getStyles('option', props)} ref={innerRef} {...innerProps}>
       <div className="flex">
-        <span className="flex self-center pr3" style={{ width: '1rem' }}>
-          {isSelected ? <IconCheck /> : null}
-        </span>
+        {isMulti ? null : (
+          <span className="flex self-center pr3" style={{ width: '1rem' }}>
+            {isSelected ? <IconCheck /> : null}
+          </span>
+        )}
         {children}
       </div>
     </div>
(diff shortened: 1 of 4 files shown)


react/components/EXPERIMENTAL_Select/Option.tsx
```diff
@@ -0,0 +1,29 @@
...
+    isMulti,
+    innerProps,
+    innerRef,
+    isSelected,
+  } = props
+  return (
+    <div style={getStyles('option', props)} ref={innerRef} {...innerProps}>
+      <div className="flex">
+        {isMulti ? null : (
+          <span className="flex self-center pr3" style={{ width: '1rem' }}>
```
igorbrasileiro, line 19: This inline style is really necessary?

react/components/EXPERIMENTAL_Select/Option.tsx
```diff
@@ -0,0 +1,29 @@
...
+            {isSelected ? <IconCheck /> : null}
+          </span>
```
igorbrasileiro, line 21: Why not
```suggestion
        {!isMulti && isSelected && (
          <span className="flex self-center pr3" style={{ width: '1rem' }}>
            <IconCheck />
          </span>
```

And add 1rem padding/margin to `children`?

vlaux: Hmmm there's a small detail that makes this not to work. The scenario is:
* multi select: i don't want anything but `{children}` 
* single select: if the option is selected, I need the check icon. Else, I need this span to have the exact same width in order to keep all options aligned in the list. Since the minimum width we have on tachyons is of 10% (not fixed btw), we cannot achieve this result. Using this fixed width as inline style avoids any possible unalignment between options.

Does it make sense?

igorbrasileiro: Yes, make sense!

react/components/EXPERIMENTAL_Select/Option.tsx
estacioneto: Did you try using `w1` from tachyons?

vlaux: Damn, I didn't know this one! Worked like a charm 👌

PR #1269, opened 2020-07-07
Fix PageHeader link button hover

vlaux:
As title says.

Button link on PageHeader has an exceeding padding on the right.
image

https://styleguide-git-fix-page-header-btn.styleguide-core.vercel.app/#/Components/Admin%20structure/PageHeader

image

(using ButtonPlain as recommended by @davicosta99)

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 
 - Table's empty state style
+- Page header button hover
 
 ## [9.124.2] - 2020-07-09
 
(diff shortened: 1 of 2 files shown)


kevinch:
YES!

estacioneto:
Me too @gustavorosolem ! What about using the `ButtonWithIcon` component? What do you think, @davicosta99 ?

jonathasbsouza:
@estacioneto @gustavorosolem we are designing mostly for desktop experiences, and the clickable area is enough for mouse clicking. Also, as we don't have other links near this Back one, the user can't misclick it, even in mobile experiences.

davicosta99:
Android and iOS increase the target size in this scenario, so I don't believe this is a problem.
https://www.w3.org/WAI/WCAG21/Understanding/target-size.html

PR #1279, opened 2020-07-14
Fix select label size on regular variation

estacioneto:

As the title says...

Regular sized select's label size should be `t-small`

##### Without the change

Input label sizes:

image

Select label sizes:

image

#### Types of changes


vlaux:
Not exactly related to this PR, but since we are fixing this label... what about fixing color too? 😬 
Just noticed that the Select's one is not using `c-on-base` 👀

PR #1290, opened 2020-07-23
Fix AutocompleteInput typing definitions

estacioneto:

As the title says.

The problem is that the `onChange` function from the `HTMLInput` is being overwritten and the current props definition doesn't express that. So, when passing the right `onChange` function, the component's types break.

##### Then

image

##### Now

image

#### Types of changes

### DIFF
--- a/react/components/AutocompleteInput/index.tsx
+++ b/react/components/AutocompleteInput/index.tsx
@@ -228,6 +228,15 @@ const AutocompleteInput: React.FunctionComponent<AutocompleteInputProps> = ({
     </div>
   )
 
+  const renderCustomMessage = (): React.ReactNode =>
+    typeof customMessage !== 'string' ? (
+      customMessage
+    ) : (
+      <div className="w-100 pa4 f6 br2 br--bottom bg-base">
+        <span className="ml3 c-on-base">{customMessage}</span>
+      </div>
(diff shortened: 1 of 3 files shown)


react/components/AutocompleteInput/index.tsx
```diff
@@ -87,7 +87,7 @@ export type AutocompleteInputProps = Omit<
   'input'
 > & {
   input: PropTypes.InferProps<typeof propTypes>['input'] &
```
vlaux, line 89: Would just changing order already solve this? So any prop we add would always overwrite the ones from input 🤔 
```suggestion
  input: React.HTMLProps<HTMLInputElement> & 
    PropTypes.InferProps<typeof propTypes>['input']
```

estacioneto: Actually the problem will still be the same. It won't override. Example:
```ts
type Props = {
  selected: boolean
} & { selected: number }
```
Gives me this:
image

estacioneto: Linked version with your suggestion:

image

estacioneto: But I'll enhance this definition to be more precise. Thanks :)

PR #1292, opened 2020-07-27
Add options.customMessage prop to AutocompleteInput

estacioneto:

As the title says, this prop should be added.

Today it's not possible to give some feedback to the users about something related to the `AutocompleteInput` options. For example, if the suggestions (loaded from the server) cannot be queried, this error can now be shown by using the `options.customMessage` prop.

There are plenty more examples of usages (error/warning messages, hints about the field, etc) and that's why the prop name is generic.

image

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- `options.customMessage` prop to `AutocompleteInput`.
+
 ### Fixed
 
 - `AutocompleteInput` typing definitions.
--- a/react/components/AutocompleteInput/README.md
+++ b/react/components/AutocompleteInput/README.md
@@ -460,7 +460,7 @@ const UsersAutocomplete = () => {
         ),
     customMessage: (
       <div className="w-100 pa4 f6 br2 br--bottom bg-base flex items-center">
-        <div className="pr3 c-muted-1 flex">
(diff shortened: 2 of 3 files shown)


estacioneto:
@JonathanAlcantara yes... I made this message to be entirely controlled by the dev who uses the `AutocompleteInput`, so almost everything the user does can make the message change if the dev wants.

About the message being shown only when loading stops, it can be controlled by the developer, as I said. I'm curious about it being bellow the spinner. Do you think it looks nicer?

image

jonathasbsouza:
It's because currently, they swap positions when options are loaded. I think the message should be fixed in the bottom (as other helping texts behave)

vlaux:
Not sure if this should be controlled by the dev or enforced by the component, but would rather go with the 2nd approach: keep options and customMessage aligned 😊 

image

estacioneto:
Oh, I see @jonathasbsouza. I'll change it then.

About that, @vlaux, I will make the example to be aligned but not enforce it by the component. There can be unknown cases that might need a different spacing. Do you agree?

vlaux:
I would say that, as default, when `customMessage` is a single string (i.e. no extra paddings/margins/etc.), it should be already aligned. In case one need any special alignment, it is always up the the dev use a component with custom styles.

estacioneto:
About the alignment, do you think this is good?
image

vlaux:
Looks good to me, but I'm not a designer, after all 😆 @jonathasbsouza 👀 ?

PR #1300, opened 2020-08-06
Exposes more props of the Dropzone component

rnbez:

Expose more properties of the `react-dropzone` lib through Styleguide's `<Dropzone />` component. This should give the community (and me =P) more flexibility when using the component.

None. It's a feature.
We're developing an admin app that only takes .xml files as the input of some of its fields. Being able to specify maximum file size and extension is extremely helpful for out app.

Clone the project and run `yarn install && yarn start`. This will launch the Styleguide page and then you can see the doc for the component as well as a working example on http://127.0.0.1:6061/#/Components/Forms/Dropzone.

styleguide-dropzone

#### Types of changes

### DIFF
--- a/react/components/Dropzone/README.md
+++ b/react/components/Dropzone/README.md
@@ -50,6 +50,60 @@ class MyDropZone extends React.Component {
 ;<MyDropZone />
 ```
 
+Size and file type restriction
+
+```js
+class MyDropZone extends React.Component {
+  constructor() {
+    super()
+    this.state = { files: null, result: null }
+    this.handleFile = this.handleFile.bind(this)
+  }
+
+  handleFile(files) {
+    this.setState({ result: files })
+  }
+
+  handleReset(files) {
+    if (files) {
+      console.log(files)
+    }
+  }
+
+  render() {
+    return (
+      <div>
+        <Dropzone
+          accept=".xml"
+          minSize={2000}
+          maxSize={10000}
+          onDropAccepted={this.handleFile}
+          onFileReset={this.handleReset}>
+          <div className="pt7">
+            <div>
+              <span className="f4">Drop here your XML or </span>
+              <span className="f4 c-link" style={{ cursor: 'pointer' }}>
+                choose a file
+              </span>
+              <p className="f6 c-muted-2 tc">Maximum file size of 10 KB.</p>
+            </div>
+          </div>
+        </Dropzone>
+        {this.state.result && (
+          <div className="mt4">
+            <p className="ttu f6">Result:</p>
+            <pre className="bg-black-025 pa4 f7">
+              {JSON.stringify(this.state.result, null, 2)}
+            </pre>
+          </div>
+        )}
+      </div>
+    )
+  }
+}
+;<MyDropZone />
+```
+
 Custom icon
 
 ```js
--- a/react/components/Dropzone/index.js
+++ b/react/components/Dropzone/index.js
@@ -53,7 +53,15 @@ class Dropzone extends PureComponent {
   }
 
   render() {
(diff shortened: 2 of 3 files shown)


react/components/Dropzone/index.js
```diff
@@ -156,6 +170,12 @@ Dropzone.propTypes = {
   icon: PropTypes.node,
   /** Allow multiple files drop */
   multiple: PropTypes.bool,
+  /** Set accepted file types. See [this MDN article](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#accept) for more information. */
+  accept: PropTypes.string,
```
rnbez, line 174: I know in the `react-dropzone` component this prop is a union of `string|string[]`, but I wanted to keep it simple here.

react/components/Dropzone/README.md
```diff
@@ -50,6 +50,60 @@ class MyDropZone extends React.Component {
...
+    if (files) {
+      console.log(files)
+    }
+  }
+
+  render() {
+    return (
+      <div>
+        <Dropzone
+          accept=".xml"
```
jonathasbsouza, line 77: Could you choose another file extension? I know that is your case, but I think would be easier for everyone to test if it was a more common extension (as .jpg, .pdf, etc)

rnbez: The code can be edited live to on the page to use any extension or mime type you want.

rnbez:
> I don't see a proper alert when the user uploads a file above the size threshold. It isn't clear what happened and how they can solve this mistake.

@jonathasbsouza The lib does not exactly tell us the reason (e.g: when props.multiple is false and multiple files are dragged and dropped.). Your _request changes_ are regarding this example not using the `onDropRejected` function with (say) an alert() inside or a more informative error for the component?

react/components/Dropzone/README.md
jonathasbsouza: > Your _request changes_ are regarding this example not using the `onDropRejected` function with (say) an alert() inside or a more informative error for the component?

Exactly! Didn't know that's a limitation of the lib =( Is there any possible workaround for this? I think it would be a bad experience for users if it's not addressed.

jonathasbsouza:
> Your request changes are regarding this example not using the onDropRejected function with (say) an alert() inside or a more informative error for the component?

@rnbez Exactly! Didn't know that's a limitation of the lib =( Is there any possible workaround for this? I think it would be a bad experience for users if it's not addressed.

jonathasbsouza:
Hey, @rnbez, I was doing some research and found [this PR](https://github.com/react-dropzone/react-dropzone/pull/938) that seems to solve this specific scenario. Can you check it, please? I'm a real noob in dev world, so I couldn't be certain about it. 😬

rnbez:
> Hey, @rnbez, I was doing some research and found [this PR](https://github.com/react-dropzone/react-dropzone/pull/938) that seems to solve this specific scenario. Can you check it, please? I'm a real noob in dev world, so I couldn't be certain about it. 😬

@jonathasbsouza that's some Sherlock stuff right there.

Here's what I did:
I read the PR you've sent and also the source code and they do pass on the list of errors that caused the file rejection. I updated the `react-dropzone` lib (`10.2.1` -> `11.0.2`) and that did it.

The problem I see is that we change the API if we update this. `onDropRejected`  will have its signatures changed from `(files: File[]) => {}` to `(fileRejections: { file: File, errors: Error[] }[])`. **This could break everyone who's already using this component.**

jonathasbsouza:
@rnbez I'll create an issue to deal with this in the next major.

PR #1308, opened 2020-08-24
fix: InputSearch icon unaligned

Thiago-Santos-SI:

Fixes ##1302.

keeps the search Icon vertically aligned, as it should be

#### How should this be tested manually?

open the InputSearch component and see if the icons are centered horizontally and vertically

#### Screenshots or usage example
before:
image

after:
image

#### Types of changes

- [x] Bug fix (an uninterrupted change that fixes a problem)
- [] New feature (a nonstop change that adds functionality)
- [] Breaking the change (fix or feature that would cause the existing functionality to be changed)
- [] Requires changes to the documentation, which has been updated accordingly.

### DIFF
--- a/react/components/InputSearch/index.js
+++ b/react/components/InputSearch/index.js
@@ -64,7 +64,6 @@ class InputSearch extends Component {
     const { disabled, size, value } = this.props
     const iconSize =
       InputSearch.iconSizes[size] || InputSearch.iconSizes.regular
-
     return (
       <Input
         {...this.props}
@@ -94,7 +93,10 @@ class InputSearch extends Component {
                   InputSearch.separatorHeight.regular,
               }}
             />
-            <span className="pointer pl4 c-link" onClick={this.handleSubmit}>
+            <span
+              className="pointer pl4 c-link"
+              onClick={this.handleSubmit}
+              style={{ height: iconSize }}>
               <SearchIcon size={iconSize} />
             </span>
           </div>
--- a/react/components/InputSearch/index.js
+++ b/react/components/InputSearch/index.js
@@ -94,9 +94,8 @@ class InputSearch extends Component {
               }}
             />
             <span
-              className="pointer pl4 c-link"
-              onClick={this.handleSubmit}
-              style={{ height: iconSize }}>
+              className="pointer pl4 c-link flex"
+              onClick={this.handleSubmit}>
               <SearchIcon size={iconSize} />
             </span>
           </div>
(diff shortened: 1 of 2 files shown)


Thiago-Santos-SI:
@JoseRenan Hi, I put it, can you check if it's right?

react/components/InputSearch/index.js
```diff
@@ -94,7 +93,10 @@ class InputSearch extends Component {
                   InputSearch.separatorHeight.regular,
               }}
             />
-            <span className="pointer pl4 c-link" onClick={this.handleSubmit}>
+            <span
+              className="pointer pl4 c-link"
+              onClick={this.handleSubmit}
+              style={{ height: iconSize }}>
```
jonathasbsouza, line 99: In `small` variation, this is not true. We need a little optical adjustment. Try using `height: 16px` instead of 14px in `small` variation and you'll see the difference.

Thiago-Santos-SI: hmm, I didn't understand what you meant, the icons are now fully centralized. 

image

estacioneto: Shouldn't `flex items-center justify-center` classes solve this problem instead of `height` prop?

jonathasbsouza: There's a difference between **math alignment** (when you get things centered exactly by their half) and **optical alignment**. In the last case, we adjust things because our eyes make wrong assumptions and think the objects aren't aligned (even if it is mathematically aligned).
In the 🔍  case, the little line below "pulls" everything down. We need to insert a 2px `padding-bottom` to "push up" and align optically.

With span height 14px and svg height 14px:
Screen Shot 2020-08-27 at 10 28 32

With span height 16px and svg height 14px:
Screen Shot 2020-08-27 at 10 28 39

Thiago-Santos-SI: ahh got it, you should have said that in the bug issue. but ok

estacioneto, line 99: What do you think?

```suggestion
              className="pointer pl4 c-link flex"
              onClick={this.handleSubmit}>
```

estacioneto: This is because we prefer to use tachyons classes instead of the `style` prop when possible :)

estacioneto (changes requested):

estacioneto (approved):

estacioneto:
Thank you again for your contribution, @Thiago-Santos-SI!
Tomorrow we will release a new version with your help in it :)

Thiago-Santos-SI:
@estacioneto cool, I didn’t know about the existence of these "tachyons classes" I thought it was a bootstrap and I looked for some class to solve it, and I didn’t, so I made this logic with the props haha.

PR #1332, opened 2020-10-01
Fixes root checkbox being always checked when all children checkboxes…

tenorius-vtex:
… are disabled

As the title says...

In the scenario of all children being disabled, as soon as the hook useCheckboxTree runs its `shake` effect, the error appears. 
This happens because the Array.prototype.every method always returns true for empty arrays ([link](https://stackoverflow.com/questions/34137250/why-does-array-prototype-every-return-true-on-an-empty-array)) making the effect mistakenly dispatch an action that checks the root checkbox 

There are two ways of handling it: Just don't do anything or if the root checkbox is checked (for some reason), uncheck it. 
I choose the latter, but I will leave it to the reviewers to decide the best approach.

[Running workspace](https://tenorio--bazarhorizonte.myvtex.com/admin/vtexlog/packing/?status=label_issued,carrier_notified)

image
#### Types of changes

### DIFF
--- a/react/components/EXPERIMENTAL_useCheckboxTree/index.tsx
+++ b/react/components/EXPERIMENTAL_useCheckboxTree/index.tsx
@@ -54,11 +54,25 @@ export default function useCheckboxTree<T>({
 
       if (!childNodes || isEmpty(childNodes)) return
 
-      const childrenChecked = childNodes
-        .filter(childNode => !isDisabled(childNode))
-        .every(childNode => checkedItems.some(comparator(childNode)))
-
       const rootChecked = checkedItems.some(comparator(tree))
+
+      const notDisabledChildNodes = childNodes.filter(
+        childNode => !isDisabled(childNode)
+      )
+
+      if (notDisabledChildNodes.length === 0) {
+        if (rootChecked) {
+          dispatch({
+            type: ActionType.Uncheck,
+            itemToToggle: { item: tree, comparator },
+          })
+        }
+        return
+      }
+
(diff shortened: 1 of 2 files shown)


react/components/EXPERIMENTAL_useCheckboxTree/index.tsx
```diff
@@ -54,12 +54,26 @@ export default function useCheckboxTree<T>({
 
       if (!childNodes || isEmpty(childNodes)) return
 
-      const childrenChecked = childNodes
-        .filter(childNode => !isDisabled(childNode))
-        .every(childNode => checkedItems.some(comparator(childNode)))
-
       const rootChecked = checkedItems.some(comparator(tree))
 
+      const notDisabledChildNodes = childNodes.filter(
```
estacioneto, line 59:
```suggestion
      const enabledChildren = childNodes.filter(
```

react/components/EXPERIMENTAL_useCheckboxTree/index.tsx
```diff
@@ -54,12 +54,26 @@ export default function useCheckboxTree<T>({
```
estacioneto, line 54: Missing CHANGELOG.md entry

estacioneto (changes requested):
I personally think that checking is different from disabling. But since this hook's logic might be completely guided using them as somehow equivalents, I think this fix can be enough.

Request changes just because of two little issues.

estacioneto (approved):

guilhermebruzzi (approved):

PR #1336, opened 2020-10-05
Fix AutocompleteInput 'double focus' when navigating with tab + arrow keys

JuliaFC:
Fix the double focus issue when navigating through the suggested options with `tab` and `arrowKeyDown`.  Also now it's possible to start over from the beginning of the suggested options when reaching the end of the list with `arrowKeyDown`

When navigating through the suggested options with `tab` and the `arrowKeyDown`, two options would be marked out with grey as shown below:
image
 
image

1. Go to [this workspace](https://tabnavigation--vtexhelp.myvtex.com/?__bindingAddress=vtexhelp.myvtex.com/)
2. Search for something (eg: Amazon)
3. If `tab` is pressed, the popover should be hidden
4. Arrow keys navigation should behave normally

Untitled-1

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 - **AutocompleteInput** restart navigating from top when reaching the end of the suggested options with the down arrow key.
->>>>>>> Update CHANGELOG.md
 
 ## [9.131.0] - 2020-10-05
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+- **AutocompleteInput** double focus issue when navigating through the suggested options with tab and arrow keys.
+
+### Added
+
+- **AutocompleteInput** restart navigating from top when reaching the end of the suggested options with the down arrow key.
+>>>>>>> Update CHANGELOG.md
+
 ## [9.131.0] - 2020-10-05
 
 ### Added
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
+
+- **AutocompleteInput** Solve double focus issue when navigating through the suggested options with tab and arrow keys.
```
lfp2, line 12: Maybe this should go on a ### Fixed title?

lfp2 (approved):
Just a nit, overall LGTM ✨

react/components/AutocompleteInput/hooks.ts
```diff
@@ -49,7 +49,7 @@ export const useArrowNavigation = (
           return setSelectedOptionIndex(index)
         case 'ArrowDown':
           e.preventDefault()
-          index = Math.min(selectedOptionIndex + 1, optionsLength - 1)
+          index = (selectedOptionIndex + 1) % optionsLength
```
estacioneto, line 52: That's interesting but when clicking the input and pressing the arrow down right away, it gives an error :(

image

estacioneto (changes requested):
That's an interesting point of view on autocomplete's accessibility :)
Requested changes because of a strange behavior :/

Klynger (approved):

CHANGELOG.md
Klynger: I agree with @lfp2

estacioneto: Also, maybe doesn't need the `Solve` word. It would sound like "Fixed AutocompleteInput double focus..."

JuliaFC: @lfp2 , @estacioneto thanks for pointing this out, I'll change it!

react/components/AutocompleteInput/hooks.ts
JuliaFC: Thank you for bringing attention to this, I'm working on fixing it, turns out it's an issue with `index = NaN` when you press the arrow down right away.

estacioneto (approved):
Nice! ✨

ThiagoAugustoSM (approved):
LGTM! ✨

PR #1338, opened 2020-10-08
Release 9.132.0

estacioneto:
### Added

- **AutocompleteInput** restart navigating from top when reaching the end of the suggested options with the down arrow key.

### Fixed

- **AutocompleteInput** double focus issue when navigating through the suggested options with tab and arrow keys.
- **EXPERIMENTAL_useCheckboxTree** root checkbox behavior when all children checkboxes are disabled.


estacioneto:
VTEX Publish failed. Need to fix this problem and then release 9.132.1

PR #1345, opened 2020-10-29
Added polyfill to fix NaN on stepper on IE11

icazevedo:

This PR fixes the behavior of the stepper component on IE11. You can see the bug [here](https://storetheme.vtex.com) (remember to use IE11).

When accessed via IE11, the stepper component previously showed `NaN` on the input field no matter which value was inserted.

This was happening because IE11 lacks support for `Number.EPSILON`. I've added a simple polyfill taken from [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/EPSILON) that fixes the problem.

You can go [here](https://icaro--storecomponents.myvtex.com) to see it working just fine (remember to use IE11).

Antes do fix: 
image

Após o fix:
image

#### Types of changes

### DIFF
--- a/react/components/NumericStepper/index.js
+++ b/react/components/NumericStepper/index.js
@@ -25,6 +25,10 @@ const validateValue = (value, min, max, defaultValue) => {
 }
 
 const formattedDisplayValue = (value, unitMultiplier, suffix) => {
+  if (Number.EPSILON === undefined) {
+    Number.EPSILON = Math.pow(2, -52)
+  }
+
   const parsedSuffix = suffix ? ` ${suffix}` : suffix
   return `${Math.round((value * unitMultiplier + Number.EPSILON) * 100) /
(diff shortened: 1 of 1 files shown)


react/components/NumericStepper/index.js
```diff
@@ -25,6 +25,10 @@ const validateValue = (value, min, max, defaultValue) => {
 }
 
 const formattedDisplayValue = (value, unitMultiplier, suffix) => {
+  if (Number.EPSILON === undefined) {
```
klzns, line 28: The fix should be adding the polyfill of EPSILON in our polyfill.io solution 

https://polyfill.io/v3/url-builder/

klzns (changes requested):

icazevedo:
The PR was closed because this is not the right place to add the polyfill. It was added to `render-server`.

PR #1358, opened 2020-12-18
Fixing font size of Input and Dropdown with small size

icazevedo:
This PR fixes a minor inconsistency between the Input and Dropdown components.  Thanks to @davicosta99 for noticing this one.

The Input and Dropdown components both had 14px font-size when their size was set to small. This caused the mobile safari app to zoom in when interacting with them. This PR sets the font-size of these components to 16px no matter their size, fixing the unwanted behavior described previously.

Open this workspace using the mobile safari app and click on the dropdown/input. It shouldn't zoom in.
[Workspace](https://icaroquantity--storecomponents.myvtex.com/fashion-eyeglasses/p)

Before and after the fix:

 

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [9.134.1] - 2020-12-22
+
 ### Fixed
 
 - Input and dropdown now have font-size of 16px no matter their sizes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
 ### Fixed
+
 - Input and dropdown now have font-size of 16px no matter their sizes
 
 ## [9.134.0] - 2020-12-04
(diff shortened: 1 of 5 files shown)


klzns (approved):
Nice gifs 💯

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- Input and dropdown now have font-size of 16px no matter their sizes
```
estacioneto, line 10: ⛏️ Just to have consistency :)
```suggestion
## [Unreleased]

### Fixed

- Input and dropdown now have font-size of 16px no matter their sizes
```

react/components/Input/index.js
```diff
@@ -154,7 +154,6 @@ class Input extends Component {
     switch (size) {
       case 'small':
         prefixSuffixGroupClasses += 'h-small '
-        classes += `${!token ? 't-small' : ''} `
```
estacioneto, line 157: So is `t-body` the default here? I see other sizes doing this 👇 
```js
classes += `${!token ? 't-body' : ''}`
```

estacioneto (approved):
Nice one! ✨ 
`CHANGELOG.md` is a "fix it then ship it" and the other one is a curiosity that might not need to be touched.

Can you review it @davicosta99 ?

react/components/Input/index.js
icazevedo: Yup, it defaults to `t-body`. Do you think it would be good to add it just for safety?

estacioneto: It can be a good thing to avoid confusion about other sizes when actually they'll be the same :)

davicosta99 (approved):
LGTM

PR #1363, opened 2021-02-08
Render the complete table when scrolling is disabled

### LINKED ISSUE #61: Tag input with Autocomplete / Multiselect

### LINKED ISSUE #1361: Can we have a parameter called fullHeight to render the table without know his size?
As we can see at the below code, we always set the table height:
https://github.com/vtex/styleguide/blob/e597cb65dc49f31b45a471839b6f470013388f28/react/components/EXPERIMENTAL_Table/Sections/index.tsx#L35

----

In this scenário, if we don't put the table height parameter, the component will show all items:

[Image from Gyazo](https://gyazo.com/a95a6ec64b3120f49037544fc194d86c)

Can We have a boolean parameter called fullHeight and if his value is true, we don't set the height?

```js
function Sections(
  { children, className, motion, disableScroll = false, fullHeight = false }: Props,
  ref: Ref<HTMLTableElement>
) {
  const { emptyState, empty, loading } = useLoadingContext()
  const { testId } = useTestingContext()
  const { tableHeight, bodyHeight } = useMeasuresContext()

  const sectionStyle = { ...motion }
  if (!fullHeight) {
    sectionStyle.height = tableHeight
  }

  return (
    <div
      style={sectionStyle}
```

matecomp:

(Resolves #1361)

With custom sections, it is possible to build lines with *dynamic height* . So, setting the table height without scrolling can hide the rendering of some items.

[Running workspace](https://mateusv2--sandboxintegracao.myvtex.com/admin/inventory/)

[Inventory](https://gyazo.com/4aa658bbbb05ec79419c84fb2c54e66b)

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 
 - **PageBlock** responsiveness issue on aside variation.
-- **EXPERIMENTAL_TableV2** rows' height when using `disableScroll`.
+- **EXPERIMENTAL_TableV2** rows height when using `disableScroll`.
 
 ## [9.135.1] - 2021-01-28
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 
 - PageBlock responsiveness issue on aside variation.
+- Render the complete table when scrolling is disabled
 
 ## [9.135.1] - 2021-01-28
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
-- PageBlock responsiveness issue on aside variation.
-- Render the complete table when scrolling is disabled
+- **PageBlock** responsiveness issue on aside variation.
+- **EXPERIMENTAL_TableV2** rows' height when using `disableScroll`.
 
 ## [9.135.1] - 2021-01-28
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+- Render the complete table when scrolling is disabled
```
estacioneto, line 12:
```suggestion
- **EXPERIMENTAL_TableV2** rows' height when using `disableScroll`.
```

estacioneto (approved):
Just a `CHANGELOG.md` suggestion, but LGTM :)

@matheusps, can you give a look?

matecomp:
When `disableScroll` is false, these styles are added to the table:
```js
'overflow-x-auto overflow-y-auto overflow-hidden'
```

Therefore, if `disabled Scroll` is true, the table does not receive any overflow style. This means that its contents go beyond the container in which it was inserted.

[Image from Gyazo](https://gyazo.com/20033afe8c79cf32b1cebf64c54d0ed2)

Should this be the expected behavior?

I believe that when we eliminate the scroll, the table should be tall enough to display all the items and keep the scroll horizontal. Makes sense?

react/components/EXPERIMENTAL_Table/Sections/index.tsx
```diff
@@ -34,7 +34,8 @@ function Sections(
     <div
       style={{ height: disableScroll ? 'auto' : tableHeight, ...motion }}
       className={classNames('mw-100', ORDER_CLASSNAMES.TABLE, {
-        'overflow-x-auto overflow-y-auto overflow-hidden': !disableScroll,
+        'overflow-x-auto overflow-y-auto': !disableScroll,
+        'overflow-x-auto': disableScroll,
```
estacioneto, line 38: Add `overflow-x-auto` to `classNames`. What do you think?

```suggestion
        'overflow-y-auto': !disableScroll,
```

PR #1368, opened 2021-02-24
Fix getChartDefaultProps ignoring the user's chosen height property

tenorius-vtex:

Allow the user to set the container's height property.

The `aspect` property in the default config will always overwrite the specified height property. Now we first analyze the user config and check if there's a height property set, and If not, then we set the default `aspect` property.

[Running workspace](https://sortbynotification--vtexlogcarriers.myvtex.com/admin/vtexlog/reports/performance/)

Ignoring the height property:
image
image
With the fix:
image
image

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [9.137.0] - 2021-03-30
+
 ### Added
 
 - **LineChart** docs warning about `container` config usage.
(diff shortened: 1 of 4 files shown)


murilovarela (commented):
Doesn't it work passing `null` to override the `customConfig.container.aspect` during the loadash's `merge`?

estacioneto (changes requested):
As we've discussed, it actually works passing `container: { aspect: null }` but it lacks documentation about the Styleguide's default aspect ratio.

Requesting this documentation here :)

estacioneto (approved):
✨

murilovarela (approved):

PR #1375, opened 2021-03-31
Fix/dangerous bulk actions experimental table

gabibguti:

Add optional `isDangerous` prop to secondary bulk actions in Experimental Table.

[Running workspace](https://gabriela--sandboxintegracao.myvtex.com/admin/received-skus/pending/)

The motivation for this change is that in Received SKUs we use the table bulk actions to approve and reject SKUs and rejecting SKUs is considered a dangerous action so it should be visually expressed in the UI.
Another motivation for this change is that the previous table had this behaviour and seems useful to other cases too.

Screen Shot 2021-03-31 at 18 39 57

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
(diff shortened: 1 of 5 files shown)


estacioneto (approved):
✨

PR #1377, opened 2021-04-20
Make date-fns dependency explicit

guilhermebruzzi:

#trivial like the title says

Explained here: https://vtex.slack.com/archives/C1MCRG7CK/p1618931965010400

"
Hi all! :worry-wave:
Styleguide is using date-fns dependency, but it seems it wasn't explicit included in the package.json because react-datepicker already includes it.
The problem with that is that if you have other dependencies other than styleguide that also depends on date-fns, than the styleguide package might not be using the one that it though it would (since it is asking for the one on the root of the node_modules, instead of being an dependency) and that can cause build problems (my case with the latest version of Cypress + latest version of Styleguide)
I made the dependency on date-fns explicit on this PR
This doesn't change any code or yarn.lock, just the package.json for those using from npm be safer.
"

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
-- Make date-fns dependency explicit so it doesn't conflict when using from npm together with other projects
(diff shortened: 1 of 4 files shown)


jonathasbsouza (approved):
LGTM

estacioneto (approved):

PR #1381, opened 2021-05-03
Fix NumericStepper and Button behavior on touch

thalytafabrine:
#### What problem is this solving?

when clicking on the add to cart or numeric stepper buttons on the mobile autocomplete, the autocomplete closes.
so the user is unable to include more items directly in the autocomplete

[Running workspace](https://thalyta--gigadigital.myvtex.com/)

#### How should this be manually tested?

- in the autocomplete, type a term and click the add to cart button of the suggested product or try to increase the quantity
- the autocomplete should not close

<details>
<summary>Add this code in <code>react/playground/Playground.tsx</code>:</summary>

```jsx
import React from 'react'
import PageHeader from '../PageHeader'
import Layout from '../Layout'
const Playground = () => (
  <Layout fullWidth pageHeader={<PageHeader title="Playground" />}>
    {/* Add your code here, don't forget to delete after */}
  </Layout>
)
export default Playground
```

</details>

#### Screenshots or example usage

https://user-images.githubusercontent.com/20840671/116931714-bed8f380-ac37-11eb-8f36-38d32ad366f4.mp4

https://user-images.githubusercontent.com/20840671/116931721-c0a2b700-ac37-11eb-84ef-4cc3525d0dce.mp4

#### Types of changes

- [x] Bug fix (a non-breaking change which fixes an issue)
- [ ] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Requires change to documentation, which has been updated accordingly.

PR #1387, opened 2021-05-13
Fix error with undefined navigator

thalytafabrine:

Fixes the error generated by the [last release](https://github.com/vtex/styleguide/pull/1381) :sadge:

image

[With error](https://thalyta1--gigadigital.myvtex.com/) - the error in this workspace is in the `AddToCartButton` which is linked with a similar logic, as I was not able to replicate the error locally for the `NumericStepper`
[Running workspace](https://thalyta--gigadigital.myvtex.com/)

<details>
<summary>Add this code in <code>react/playground/Playground.tsx</code>:</summary>

```jsx
import React from 'react'
import PageHeader from '../PageHeader'
import Layout from '../Layout'
const Playground = () => (
  <Layout fullWidth pageHeader={<PageHeader title="Playground" />}>
    {/* Add your code here, don't forget to delete after */}
  </Layout>
)
export default Playground
```

</details>

#### Types of changes

### DIFF
--- a/react/components/NumericStepper/index.js
+++ b/react/components/NumericStepper/index.js
@@ -285,8 +285,8 @@ class NumericStepper extends Component {
 
     const touchDevice =
       'ontouchstart' in window ||
-      navigator.maxTouchPoints ||
-      navigator.msMaxTouchPoints
+      (typeof navigator !== 'undefined' &&
+        (navigator?.maxTouchPoints || navigator?.msMaxTouchPoints))
 
     const content = (
       <React.Fragment>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+- Error with `undefined` navigator.
+
 ## [9.141.0] - 2021-05-13
 
 ### Fixed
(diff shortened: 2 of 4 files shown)


react/components/NumericStepper/index.js
```diff
@@ -284,9 +284,11 @@ class NumericStepper extends Component {
...
 
     const touchDevice =
-      'ontouchstart' in window ||
-      navigator.maxTouchPoints ||
-      navigator.msMaxTouchPoints
+    'ontouchstart' in window
+      ? true
+      : typeof navigator === 'undefined'
+      ? false
+      : navigator?.maxTouchPoints || navigator?.msMaxTouchPoints
```
estacioneto, line 291: This can be simpler, right? I've put the `navigator != null` just because I'm not 100% sure about the support of the optional chaining. If we transpile it correctly on JS, then it's better to use it.

```suggestion
    'ontouchstart' in window || (navigator != null && (navigator.maxTouchPoints || navigator.msMaxTouchPoints))
```

estacioneto (changes requested):

react/components/NumericStepper/index.js
thalytafabrine: unfortunately that doesn't work. the error occurs when trying to access the value of the variable when it doesn't exist, so I need to use typeof 😞

estacioneto: Oh, I see... Maybe this, then? 👇 
```suggestion
    'ontouchstart' in window || (typeof navigator !== 'undefined' && (navigator.maxTouchPoints || navigator.msMaxTouchPoints))
```

thalytafabrine: done!

estacioneto (approved):

PR #1390, opened 2021-05-31
Fixing window scroll not defined

razvanudream:

This is related to the fix done here https://github.com/vtex/styleguide/pull/1371 - if you're installing the current version of the app on a workspace where Modal is used, on hard refresh (CTRL + Shift + R) the window.scroll stays undefined during componentDidUpdate.  For some reason the file is not imported properly on hard refresh, after moving the function to Modal/index.js and calling it right before the logic in ComponentDidUpdate the problem was fixed

Can't provide a workspace link because the functionality requires for you to have an account and some dummy data inserted in master data for your email in order for the contracts to show up in your account, but here's a video that describes the issue and shows the fix.

https://user-images.githubusercontent.com/71461884/120172465-c1654380-c20b-11eb-848b-dd2c3a6beb39.mp4

1. Install the latest version of styleguide on a workspace that uses Modal and hard refresh on that page to get the error 
2. Link this fixed version on the same workspace and repeat 1) -> the error will no longer appear

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+- **Modal** sometimes not loading `window.scroll` polyfill after the browser hard refresh.
+
 ## [9.142.0] - 2021-05-19
 
 ### Added
--- a/react/components/Modal/index.js
+++ b/react/components/Modal/index.js
@@ -8,8 +8,13 @@ import TopBar from './TopBar'
 import BottomBar from './BottomBar'
 import styles from './modal.css'
 
-import '../../modules/scrollPollyfill'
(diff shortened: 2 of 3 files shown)


estacioneto:
Hi, @razvanudream!

Thank you for your contribution. It's a bizarre problem, right? But even so, I think the solution is fine for us to ship.

Could you please add this to the changelog under the `[Unreleased]` section? 👇 

```md
### Fixed

- **Modal** sometimes not loading `window.scroll` polyfill after the browser hard refresh.
```

razvanudream:
Hi @estacioneto - thanks for the review! It is a bit of a thinker this one, but I didn't want to spend too much time to try and understand it, I just blamed it on the import and noticed that moving the function within the same file worked.   

Updated the CHANGELOG just now.

Cheers,

Razvan

estacioneto (approved):

PR #1396, opened 2021-06-23
Add handle callback on toggle fields visibility

wagnerlduarte:

Possibility to get callback on toggle fields visibility with `onToggleColumn`, `onHideAllColumns` and `onShowAllColumns`
The callback functions `onHideAllColumns` and `onShowAllColumns` returns a string array with active columns, and `onToggleColumn` returns the field changed and a string array with active columns.

I needed to create a feature that save on localstorage a custom visibility created by user, but the controls handlers are not exposed to get a callback when the user did the visibility action.
Because of this I added 3 functions to get this actions and a callback with active fields, to solve my problem and for everyone who wants to use this callback to do something.

```jsx
toolbar={{
...
          fields: {
             ...
            onToggleColumn: (params) => {
              console.log(params.toggledField)
              console.log(params.activeFields)
            },
            onHideAllColumns: (activeFields) => console.log(activeFields),
            onShowAllColumns: (activeFields) => console.log(activeFields),
            ...
          },
...
}}
```

image

#### Types of changes

### DIFF
--- a/react/components/Table/index.js
+++ b/react/components/Table/index.js
@@ -90,8 +90,9 @@ class Table extends PureComponent {
       newFieldsArray
     )
 
-    toolbar?.fields?.onToggleColumn &&
+    if (toolbar?.fields?.onToggleColumn) {
       toolbar.fields.onToggleColumn({ toggledField: key, activeFields })
+    }
   }
 
   handleShowAllColumns = () => {
--- a/react/components/Table/index.js
+++ b/react/components/Table/index.js
@@ -4,6 +4,7 @@ import PropTypes from 'prop-types'
 import reduce from 'lodash/reduce'
 import map from 'lodash/map'
 import isEmpty from 'lodash/isEmpty'
+import difference from 'lodash/difference'
 
 import Box from '../Box'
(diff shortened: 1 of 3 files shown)


react/components/Table/index.js
```diff
@@ -81,14 +82,35 @@ class Table extends PureComponent {
...
+
+    const { toolbar } = this.props
+
+    const activeFields = difference(
+      Object.keys(this.props.schema.properties),
+      newFieldsArray
+    )
+
+    toolbar?.fields?.onToggleColumn &&
+      toolbar.fields.onToggleColumn({ toggledField: key, activeFields })
```
estacioneto, line 94: 💡 Just an opinion

Since you're not returning any value, maybe an `if` statement would be better, don't you think?
```suggestion
    if (toolbar?.fields?.onToggleColumn) {
      toolbar.fields.onToggleColumn({ toggledField: key, activeFields })
    }
```
The same applies to other functions

estacioneto (approved):
Just a tip, so it's a fix it then ship it :)

PR #1398, opened 2021-07-01
Remove form element to avoid dom validation error

salesfelipe:

- When the InputSearch is used inside a <form> the react validation gives you this error.

[Running workspace](https://cabaceiras--recorrenciacharlie.myvtex.com/admin/app/subscriptions/plans/vtex.subscription.testinho/)

Screen Shot 2021-07-01 at 5 08 32 PM

<details>
<summary>Add this code in <code>react/playground/Playground.tsx</code>:</summary>

```jsx
import React from 'react'
import PageHeader from '../PageHeader'
import Layout from '../Layout'
const Playground = () => (
  <Layout fullWidth pageHeader={<PageHeader title="Playground" />}>
    {/* Add your code here, don't forget to delete after */}
  </Layout>
)
export default Playground
```

</details>

#### Types of changes


lbebber (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+- Remove **form** element from **InputSearch** to avoid errors of dom validation.
```
estacioneto, line 12:
```suggestion
### Removed

- **EXPERIMENTAL_TableV2** `form` element from search input to avoid DOM validation errors. 
```

estacioneto (changes requested):
Requesting changes just because of `CHANGELOG.md`.

@matheusps do you recall if there's any special reason to have a `form` element here? I saw you were the one to add it here

matheusps:
@estacioneto, I've added the form because the "search input" is never just an input. The idea was to make a search form, like: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Search_role (it's missing the role, btw).

salesfelipe:
Perhaps a prop `shouldUseForm`?  It defaults to true, what do you guys think @estacioneto , @matheusps ?

estacioneto:
Interesting, @matheusps. @salesfelipe, have you tried to add the role or the error persist? Got curious now.

And about the prop, I'm not entirely sure. Since we don't add an `onSubmit` handler to the form, maybe it shouldn't really be there.

estacioneto (approved):
Since there was no response, I'll approve this PR and release a version as soon as possible :)

matheusps:
Just adding my two cents here, maybe a little late. I think the `shouldUseForm` can be a nice approach here. Since @salesfelipe is using formik, he doesn't need this form - other users may. I would also add the submit to the form to fix the behavior.

estacioneto:
Oh, I see, @matheusps. 

It's up to you, @salesfelipe. If you'd like to add the prop and fix the form behavior too, we can make a different release in the same PR.

PR #1427, opened 2022-08-30
Fix: `NumericStepper` click event

eduardoformiga:

This PR aims to solve  [this Known Issue (KI)](https://vtex-dev.atlassian.net/browse/SF-35):
[KI] Product Quantity Selector Error for high-resolution screens.

Copied from Jira Issue:
When rendering high resolutions, the quantity selector doesn't work.

Simulation
 - Access the `storecomponents`;
 - Open the Chrome inspector
 - Set the screen size on 4000px X 2000px with the chrome's DevTools utility
 - Refresh the page;
 - Wait for the page to be loaded;
 - Close the inspector  WITHOUT REFRESHING THE PAGE;
 - Try to use the qty selector.

---

While investigating this issue, we noticed that this problem is not related to the high-resolution screens, but to the toggle device toolbar feature of devtools that simulates some touch devices.

Screen Shot 2022-08-30 at 12 35 13

As we also have a [previous fix regarding](https://github.com/vtex/styleguide/pull/1381) the quantity selector inside the `autocomplete` component, we understand that when we use touch devices, the click event on quantity selector button is handled by the `onTouchEnd` event, instead of `onClick`, but the values that comes from the older `touchDevice` is not updated once we switch to touch devices.

For the sake of simplifying this, we are only handling the event using the `onClick` for all kinds of devices, and we are using the `event.stopPropagation` to keep the https://github.com/vtex/styleguide/pull/1381 working as expected.

|Before|After|
|-|-|
|xUAtuZtIuh|ow68BiiYts|

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+- **NumericStepper** not working after switching from touch devices.
+
 ## [9.146.1] - 2022-03-14
 
 ### Removed
--- a/manifest.json
+++ b/manifest.json
@@ -2,7 +2,7 @@
   "vendor": "vtex",
   "name": "styleguide",
   "title": "VTEX Styleguide",
-  "version": "9.146.1",
+  "version": "9.146.2",
   "description": "The VTEX Styleguide components for the Render framework",
   "builders": {
     "react": "3.x"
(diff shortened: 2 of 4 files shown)


danzanzini (approved):
Works great. I couldn't test the behavior at the autocomplete. How did you test it?

thalytafabrine (approved):

eduardoformiga:
> 

I tried so hard to test it using the `gigadigital` account but it's inactive, and also the `ametllerorigen` that has this specific scenario of `QuantitySelector` inside the `Autocomplete` component but I bumped into the regionalization API bringing some errors inside my custom workspace. Since it is a very specific case, I will look into this behavior after publishing.

